PR #5, opened 2026-04-03
feat: use tmux display-popup for in-UI session attach

lucascaro:
## Summary

When using the tmux backend, pressing Enter on a session no longer quits and restarts the TUI. The TUI now suspends in place via `tea.ExecProcess` while the session is attached. When you detach, the TUI resumes immediately with all state intact.

### On tmux ≥ 3.2 (running inside a tmux session)

The session opens as a **floating `display-popup` overlay** (95% × 90%) that sits on top of the TUI. Closing the popup (or the agent exiting) returns you directly to the running TUI — no full-screen switch, no restart.

### On older tmux or outside a tmux session

Falls back to a plain full-screen `tmux attach-session`, still with the no-restart benefit since `tea.ExecProcess` resumes the TUI after detach.

### Native PTY backend

Unchanged — continues to use the existing quit+restart loop in `cmd/start.go`.

## Changes

| File | What |
|------|------|
| `internal/tmux/version.go` | New `Version()` / `SupportsDisplayPopup()` helpers |
| `internal/mux/interface.go` | Added `SupportsPopup()`, `UseExecAttach()`, `PopupAttach()` to `Backend` interface |
| `internal/mux/tmux/backend.go` | Implements the three new methods; `PopupAttach` runs `tmux display-popup -E -w 95% -h 90% -- tmux attach-session -t <target>` |
| `internal/mux/native/backend_{unix,windows}.go` | Stub implementations (`SupportsPopup=false`, `UseExecAttach=false`) |
| `internal/tui/messages.go` | Added `AttachDoneMsg{Err, RestoreGridMode}` |
| `internal/tui/app.go` | New `doAttach()` helper; all attach paths call it; `AttachDoneMsg` handler refreshes preview on return |
| `cmd/start.go` | Clarified the quit+restart loop is native-backend-only |
| `CHANGELOG.md`, `docs/features.md` | Updated |

## Testing

All existing tests pass. Verified manually — popup opens, interaction works, TUI resumes cleanly on detach.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   row) has been merged into one line — status dot, agent badge, session title, project
   name, and worktree badge all appear on a single header line, giving each cell one
   extra row of terminal-output preview.
+- **Session attach via in-process overlay**: when using the tmux backend, pressing
+  Enter on a session no longer quits and restarts the TUI. Instead the TUI suspends
+  in place using `tea.ExecProcess` while the session is attached. When you detach,
+  the TUI resumes immediately with all state intact — no disk reload, no flicker.
+  - On **tmux ≥ 3.2** the session opens as a floating `tmux display-popup` overlay
+    (95 % width, 90 % height) that sits on top of the TUI. Closing the popup (or the
+    agent exiting) returns you directly to the TUI without any full-screen switch.
+  - On older tmux or when running outside a tmux session, a plain full-screen
+    `tmux attach-session` is used, still with the no-restart improvement.
+  - The **native PTY backend** is unaffected and continues to use the existing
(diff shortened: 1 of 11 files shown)


lucascaro:
All review comments addressed:

- **Evaluation order bug** (3 instances, lines 298/395/1960): Fixed — all now use `cmd := m.doAttach(...); return m, cmd` pattern
- **Shell dependency**: Removed `sh -c` wrapper; now calls `exec.Command("tmux", ...)` directly
- **`PopupAttach` nil guard**: Added `active == nil` check returning a descriptive error
- **`SessionDetachedMsg` comment**: Updated to clearly state this is legacy native-backend-only path
- **`Version()` tests**: Factored parsing into pure `parseVersion()` + added `version_test.go` covering normal formats, letter suffixes, error cases, and `supportsDisplayPopup` boundary values

PR #11, opened 2026-04-03
fix: prevent tests from touching real ~/.config/hive

lucascaro:
## Summary
- Add nil guard in `saveState()` so nil `Projects` writes `[]` instead of `"null"`, preventing cascading state corruption
- Replace eager `init()` loggers in 4 files with lazy `sync.Once` initialization — avoids creating/opening `~/.config/hive/hive.log` at package load time before tests can override `HIVE_CONFIG_DIR`
- Add `TestMain` in the `tui` package that sets `HIVE_CONFIG_DIR` to a temp dir, ensuring `New()` and other runtime code never resolves against the real config dir during tests

## Test plan
- [x] All existing tests pass (`go test ./... -count=1`)
- [x] New test `TestSaveState_NilProjectsWritesEmptyArray` verifies the nil guard
- [ ] Run tests and verify `~/.config/hive/state.json` mtime is unchanged

### DIFF
--- a/internal/tui/app.go
+++ b/internal/tui/app.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/charmbracelet/bubbles/key"
@@ -25,15 +26,19 @@ import (
 	"github.com/lucascaro/hive/internal/tui/styles"
 )
 
-var debugLog *log.Logger
+var (
+	debugLog     = log.New(os.Stderr, "[hive] ", log.Ltime)
+	debugLogOnce sync.Once
+)
 
-func init() {
-	f, err := os.OpenFile(config.LogPath(), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o600)
-	if err != nil {
-		debugLog = log.New(os.Stderr, "[hive] ", log.Ltime)
-		return
-	}
-	debugLog = log.New(f, "[hive] ", log.Ltime|log.Lmicroseconds)
+func initDebugLog() {
+	debugLogOnce.Do(func() {
+		f, err := os.OpenFile(config.LogPath(), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o600)
+		if err != nil {
+			return
+		}
+		debugLog = log.New(f, "[hive] ", log.Ltime|log.Lmicroseconds)
+	})
 }
 
 // Model is the root Bubble Tea model.
@@ -89,6 +94,10 @@ func (m Model) LastAttach() *SessionAttachMsg { return m.attachPending }
 
 // New creates the root model.
 func New(cfg config.Config, appState state.AppState) Model {
+	initDebugLog()
+	components.InitSidebarLog()
+	components.InitStatusLog()
+	components.InitPreviewLog()
 	km := NewKeyMap(cfg.Keybindings)
 	ni := textinput.New()
 	ni.CharLimit = 256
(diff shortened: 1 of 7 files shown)


lucascaro (commented):
## Review of PR #11

Overall this is a solid fix. The lazy logger pattern and TestMain approach are the right solutions. A few observations:

### 1. Nil guard in saveState() -- Correct

The nil guard in `internal/tui/persist.go` is correct. `json.MarshalIndent(nil, ...)` produces `"null"` which would corrupt `state.json`, while `json.MarshalIndent([]*state.Project{}, ...)` produces `"[]"`. The test expectation matches.

### 2. sync.Once + package var reassignment -- Safe but worth noting

The pattern of initializing `debugLog` to a stderr fallback at var-declaration time, then reassigning inside `sync.Once`, is safe. `sync.Once` guarantees the closure completes before any concurrent `Do` call returns, so no goroutine can observe a partially-written `*log.Logger` pointer. The `log.Logger` methods themselves are goroutine-safe.

One thing to note: the file handle opened inside `debugLogOnce.Do(...)` is never closed (same for sidebar/status/preview loggers). This was also true with the old `init()` approach, so it's not a regression -- just a pre-existing leak that lives for the process lifetime. Fine in practice.

### 3. TestMain vs t.Setenv in persist_test.go -- Potential interaction

This is the most subtle point. `TestMain` in `internal/tui/app_test.go` calls `os.Setenv("HIVE_CONFIG_DIR", dir)` at the process level. Then `persist_test.go`'s `setHomePersist()` calls `t.Setenv("HIVE_CONFIG_DIR", ...)` which will:
1. Save the current value (the TestMain temp dir)
2. Override it with the test-specific path
3. Restore the TestMain value on test cleanup

This is actually correct behavior -- persist tests get their own isolated dirs, and the TestMain value acts as a safety net for any test that does NOT call `setHomePersist()` (e.g., `TestSaveState_NilProjectsWritesEmptyArray` does call it via `setHomePersist`). So this is fine. No conflict.

### 4. Missed init() or test files

**cmd/ init() functions** (`cmd/root.go`, `cmd/start.go`, `cmd/mux_daemon.go`) -- these are cobra flag registration only, no file I/O. Safe to leave as-is.

**components test files** -- there are 7 test files in `internal/tui/components/` (`agentpicker_test.go`, `confirm_test.go`, `gridview_test.go`, `preview_test.go`, `sidebar_test.go`, `statusbar_test.go`, `titleedit_test.go`). These are in Go package `components`, not `tui`, so the `TestMain` in `app_test.go` does NOT apply to them. However, since the PR converts the `init()` loggers to lazy initialization, the `components` tests will no longer trigger file creation at import time -- the loggers just default to stderr. I confirmed none of these test files reference `config.LogPath`, `config.Dir`, or `config.StatePath`, so they're fine today.

**However**: if a future test in the `components` package calls code that triggers `InitSidebarLog()` / `InitStatusLog()` / `InitPreviewLog()`, it would hit the real `~/.config/hive/hive.log`. You may want to add a `TestMain` to `internal/tui/components/` as well for defense-in-depth, but it's not strictly needed for this PR.

### 5. debugLog used before initDebugLog() -- Safe

`debugLog` is initialized at var-declaration time to `log.New(os.Stderr, "[hive] ", log.Ltime)`, so it is never nil. If `saveState()` is somehow called before `New()` (e.g., in tests), the log just goes to stderr instead of the file. This is the correct fallback. Same applies to the component loggers.

### Verdict

Looks good. The approach is clean and the edge cases are handled. Ship it.

PR #92, opened 2026-04-13
feat(grid): extend cell above empty space to fill screen

### LINKED ISSUE #89: Extend grid cells to fill empty space in grid view
## Description

In grid mode, when the number of sessions does not evenly fill the grid layout, empty cells are left at the bottom of the screen. Instead of leaving blank space, the cell(s) directly above the empty space should be extended vertically to fill the remaining screen area. This ensures the full terminal height is always utilized and improves the visual density of the grid layout.

lucascaro:
## Summary

- Switches grid rendering from row-first to column-first so each column can independently control cell heights
- Columns with no content in the last row extend their last real cell downward to absorb the empty slot
- Fully-occupied columns get the remaining height (after integer division) in their last cell

Fixes #89

## Test plan

- [ ] Verify with an odd number of sessions (e.g. 3, 5, 7) that the cell above the empty slot extends to fill the screen
- [ ] Verify with an even/fully-filled grid that layout is unchanged
- [ ] `go test ./...` passes (existing height-invariant tests + new `TestGridView_LastRowFillsHeight`)


lucascaro (commented):
Self-review findings addressed:

**BLOCKING fixed:** `CellAt()` was still using uniform `cellH` for all rows. With variable-height last rows, mouse clicks in the extended portion of a sparse column's last cell were mapped to a non-existent row and silently dropped. Fixed in the latest commit — `CellAt` now mirrors `View()`'s height logic with the same `lastCellH` / `emptyInLastRow` derivation.

**Test coverage note:** `TestGridView_LastRowFillsHeight` verifies the height invariant and that all sessions render, but doesn't assert the extended cell height directly. The existing `TestGridView_ExactHeight_VariousCounts` covers the invariant across 1–9 sessions; direct height-per-column assertion would require parsing ANSI output which is brittle. The behavior is covered by the height invariant tests and the user-reported regression.

Issue #112, opened 2026-04-15
Make all key bindings configurable, consistent, and documented


lucascaro:
## Description

Several key bindings in hive are hard-coded and cannot be customized by users — for example, `Enter` to attach to a session and `i` to enter interact mode. Some of these keys are also undocumented, making them discoverable only by reading source code. This inconsistency means users cannot tailor hive to their workflow, and the keymap documentation is incomplete.

All key bindings must follow consistent patterns, live in a single source of truth, be user-configurable (via settings/config), and be documented (in the help panel and user-facing docs).

## Acceptance Criteria

- All key bindings are defined in one place (no hard-coded keys scattered across handlers).
- Every binding is configurable through the standard config/settings mechanism.
- Every binding appears in the help panel and relevant user-facing documentation.
- Going forward, adding or changing any key binding **requires** updating both the configuration surface and the documentation — this becomes part of the standard AC for key-related changes.

lucascaro:
Completed — keybindings are fully configurable, settings UI exposes all actions, docs updated, policy added to AGENTS.md. PR #116 + incremental commits.

PR #114, opened 2026-04-15
refactor(config): keybindings []string foundation (chunk 1/3 of #112)

lucascaro:
## Summary

First of three chunks for #112 (make all keybindings configurable, consistent, and documented). This chunk is foundation only — **no behavior changes, no help-overlay changes, no flow tests modified**.

- Convert every `KeybindingsConfig` field from `string` to a new `KeyBinding` (`[]string`) type so users can bind multiple keys to one action.
- `KeyBinding.UnmarshalJSON` accepts both the legacy single-string form (`"a"`) and the new array form (`["up","k"]`), so existing `~/.config/hive/config.json` files load unchanged. Marshalling always produces the array form, so the next save migrates the on-disk shape.
- Bump `SchemaVersion` 4 → 5. Migration uses reflection to fill any empty `KeyBinding` from defaults, so newly-added action fields (`Detach`, `CursorUp/Down/Left/Right`, `SessionColorNext/Prev`, `ToggleAll`, `InputMode`, `CollapseItem`, `ExpandItem`) populate without per-field code.
- `tui/keys.go` and the settings UI's `keybindField` helper updated to operate on the new slice type. Help-overlay labels preserved exactly (`↑`, `←/h`, `space`, etc.) so the existing golden snapshots pass unmodified.

## What's deferred

- **Chunk 2:** centralize handlers — replace literal `msg.String()` switch cases in `handle_keys.go`, `gridview.go`, pickers, help.go with `key.Matches` against the configured KeyMap. Plumb KeyMap into components. Replace hard-coded `ctrl+q` at `gridview.go:227` with `km.Detach`.
- **Chunk 3:** default behavior change (`Enter` → input mode, `f` → attach), drop `tab`/`i` defaults, settings UI fields for new actions, `docs/keybindings.md` rewrite, `AGENTS.md` keybindings policy, ~70 flow-test updates.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` — all packages green; new tests for `KeyBinding` JSON round-trip, v4→v5 migration fill, and user-customization preservation.
- [ ] Manual: load an existing `config.json` written under schema 4 — should load and re-save as `["a"]` form without losing user customizations.


lucascaro:
Pushed `a39afa6` addressing the review findings:

**Blocking**
- Reverted `MoveUp/Down/Left/Right` defaults to single-key (`shift+arrow`) — capital-letter aliases moved to chunk 2 where they'll land alongside the explicit handler refactor.
- Tightened CHANGELOG so the new schema fields aren't claimed as "configurable" yet (they're schema-only in this release).

**Important**
- `KeyBinding.MarshalJSON` now emits `[]` for nil instead of `null`, with a regression test (`TestKeyBinding_MarshalNilEmitsEmptyArray`).
- `keybindField` description now surfaces the multi-value comma syntax to users.
- `bindLabeled` collapsed into `bind(kb, label, desc, extras...)` — empty `label` triggers auto-join via `strings.Join`.

**Minor**
- `mergeKeybindingDefaults` got a comment explaining why reflection is justified here vs the explicit per-field style of other migrations.
- Dropped the empty `if SchemaVersion < 5 { }` block; rationale moved beside the actual fill site.

`go build`, `go vet`, `go test ./...` all green.

PR #212, opened 2026-05-23
feat(gui): keyboard copy/paste/select-all that survives mouse tracking

qmertesdorf:
## Summary

- When an inner program (Claude CLI, vim, htop) enables DEC mouse tracking, xterm.js forwards mouse events to the PTY and selection becomes impossible — leaving no way to copy assistant output from a Hive tile.
- Add `Ctrl+Shift+A` (select all scrollback), `Ctrl+Shift+C` (copy selection), `Ctrl+Shift+V` (paste from clipboard) via `attachCustomKeyEventHandler` in `cmd/hivegui/frontend/src/main.js`, so the shortcuts are intercepted before xterm forwards them to the PTY.
- Conventions match VS Code's integrated terminal and Windows Terminal. Clipboard calls are best-effort (`.catch(() => {})`) so a denied permission prompt does not break the terminal.

## Motivation

Reproducer on Windows: open a session running Claude CLI, try to mouse-select any output — the inner program eats the mousedown, no selection appears. There is currently no other path to copy text out of a tile short of reading the JSONL transcript from disk.

## Notes

- Does not touch the existing click-to-position / linkifier / mouse-down handlers above this block; those bail out early when modifier keys are held, so there is no interaction.
- No new dependencies. ~29 LOC, single file.

## Follow-up (separate PR)

A true `Shift+drag` bypass that lets the user make a partial mouse selection while mouse tracking is active. Bigger change — requires intercepting mouse events in capture phase and driving `term.select()` manually.

## Test plan

- [ ] Build Windows target: `bash build.sh --platform windows`
- [ ] Launch Hive, start a Claude CLI session
- [ ] `Ctrl+Shift+A` selects all visible scrollback
- [ ] `Ctrl+Shift+C` copies; paste into a text editor and verify content
- [ ] `Ctrl+Shift+V` pastes clipboard contents into the prompt
- [ ] Sanity check that the same shortcuts in a non-mouse-tracking shell (e.g. a plain `bash` prompt) still behave correctly and don't double-paste

### DIFF
--- a/cmd/hivegui/app.go
+++ b/cmd/hivegui/app.go
@@ -13,6 +13,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/atotto/clipboard"
 	"github.com/lucascaro/hive/internal/agent"
 	"github.com/lucascaro/hive/internal/buildinfo"
 	hdaemon "github.com/lucascaro/hive/internal/daemon"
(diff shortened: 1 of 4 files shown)


qmertesdorf:
Pushed a follow-up fix (`566fc84`): the copy/paste/select-all handler in this PR was registered via a second `attachCustomKeyEventHandler()` call, but xterm.js keeps only **one** custom key handler — the later registration (Cmd+Backspace / Ctrl+`) silently replaced it, so Ctrl+Shift+C/V/A did nothing at runtime. Merged both into a single handler, with the Cmd+Backspace and Ctrl+` checks ordered above the Ctrl+Shift guard so all four bindings survive. Verified live on a Windows build (copy now works) and all 91 frontend unit tests pass.

lucascaro (approved):
Reviewed: macOS flows (mouse selection, Cmd+C, Cmd+Backspace→Ctrl+U, Ctrl+`) are not regressed — the new Ctrl+Shift+C/V/A guard excludes metaKey so Cmd-based copy still works natively. Bindings are cross-platform (additive on Mac, no conflict). Approving.

PR #277, opened 2026-08-23
feat(worktree): browse, resume and clean up worktrees per project

lucascaro:
## Why

A worktree could not outlive its sessions. `registry.Kill` removed it when the last session in it closed, and `ReclaimOrphanWorktrees` force-deleted every unclaimed `.worktrees/*` directory at daemon startup. Nothing ever called `git worktree list` or `git branch`, and `worktree.Cleanup` left `refs/heads/<branch>` behind forever.

The result: work got stranded. You could not see what worktrees existed, could not resume one, and orphaned branches accumulated invisibly.

## What changed

**Lifecycle.** A worktree now survives its session unless it is *pristine* — no uncommitted changes, no unpushed commits, **and** a resolvable comparison base. An unanswerable base counts as "holds work", never as "safe to delete"; the conservative direction is load-bearing, because the other reading silently destroys commits. Pristine worktrees are still pruned on close and at startup, so throwaway sessions leave nothing behind.

**A per-project browser** (`⌘E`, the ⎇ button on a project row, or the command palette) lists every worktree with what it holds — occupying sessions, uncommitted changes, unpushed commit count — plus the local branches that have no worktree. From it you can:

- start a session in an existing worktree to pick that work back up
- create a worktree for an orphaned branch
- rename a worktree (branch and directory stay coupled)
- delete a worktree, or delete an orphaned branch

**Deleting is deliberately hard to do by accident.** Refused outright while a session is running inside (force does **not** override). Uncommitted changes, unpushed commits, or an unmerged branch each require a confirmation that names exactly what would be lost. The branch is kept by default so commits stay recoverable; removing it too is a separate opt-in.

`⌘T`'s worktree checkbox now also accepts a branch name; empty keeps the existing auto-generated adjective-noun.

## Shape

Bottom-up, following the one-way `wire → worktree → registry → daemon` dependency:

| Layer | Change |
|---|---|
| `internal/worktree/inventory.go` | `List`, `ListBranches`, `Inspect`/`Status.Pristine`, `MainRoot`, `MoveWorktree`, `RenameBranch`, `DeleteBranch` |
| `internal/registry/worktrees.go` | `ListWorktrees`, `RemoveWorktree`, `CreateWorktreeForBranch`, `RenameWorktree`, `DeleteBranch` + refusal sentinels |
| `internal/registry` (existing) | `Kill` and the startup reclaim prune only pristine worktrees; `Create` adopts an existing worktree so a resumed one stays claimed |
| `internal/wire` | frames `0x16`–`0x1b`, `CreateSpec.WorktreePath`, three new error codes |
| clients | `cmd/hivegui`, `cmd/hived-ws-bridge`, `internal/wire/testclient` updated in lock-step |

`git worktree list` stays the single source of truth — the registry stores no worktree records of its own.

## UX fallout worth reviewing

Prompts that risk losing work no longer use a native OS alert. `Confirm()` can only answer yes/no, which forced the branch question and the deletion question into two sequential prompts with no way to back out of the second. They now share one in-DOM dialog that states the stakes and offers the real outcomes as buttons.

Building that surfaced focus bugs the **older** modals already had, so the fixes are in shared code and they inherit them:

- Tab escaped every dialog into the page behind it — where the next stop is a hidden terminal's textarea, so keystrokes went somewhere invisible. Three hand-rolled focus traps (help overlay, settings, dialog) are now one tested helper, and the worktree panel had none at all.
- Closing a modal could strand focus on `<body>`, or leave it inside the just-hidden modal.
- `Escape` inside an inline rename closed the surrounding panel and discarded the edit — `keyboard.ts` listens in the **capture** phase, so the input's `stopPropagation` could never win that race.
- The worktree rename re-implemented `beginInlineRename`; it now uses it.

## Behaviour change

Force-closing a session with uncommitted changes no longer destroys its worktree. Intentional, and called out in `CHANGELOG.md` under `Changed`.

## Tests

- **Go** — every refusal path (in-use / dirty / unpushed / unmerged / path containment), the pristine-prune lifecycle change, adoption-keeps-the-claim, rename rollback, wire round-trips per golden principle 6, and daemon dispatch. 85.7% / 81.3% statement coverage in `worktree` / `registry`.
- **Frontend** — 514 unit + DOM, 172 mock e2e including a focus-containment sweep over *every* modal, 8 real-daemon e2e against real `git`.
- Fixes were each verified to fail without the patch; the tests that could not fail were rewritten.

`e2e-real` now points `hived` at a throwaway repo via `--cwd`. Without it the harness adopted the real hive checkout as its default project, and these specs would have created and deleted worktrees in the developer's own clone.

## Verification

```
go vet ./... && go test ./... -count=1
cd cmd/hivegui/frontend && npx tsc --noEmit && npx biome ci .
npm test && npx playwright test && npm run test:e2e:real
wails build
```

All green. The three `scroll-codex` / `scroll-restream-strand` failures in `e2e-real` are the pre-existing flakes documented in `docs/product-specs/245`; I confirmed they fail identically with this branch's harness change reverted.


lucascaro:
## Autofix Summary

Review findings addressed in 00a9f87. One was a genuine data-loss bug, reproduced with a test before fixing.

**BLOCKING — closing a session could delete your working directory.** When a project's cwd is itself a linked worktree (how hive is run from inside its own `.worktrees/`), every plain session claimed that directory as hive-managed, and `Kill` deletes what it claims. Reproduced: a file committed *and merged* was destroyed on session close. Introduced by `adoptDetachedWorktree` in this PR; no tagged release shipped it.

Fixed in two independent layers:
- Adoption is restricted to `<mainRoot>/.worktrees/` and never takes the project's own cwd, even when that sits under `.worktrees/`.
- `Kill` re-checks containment itself rather than trusting `Entry.WorktreePath` — teardown is the last place an upstream bug can still cost someone their checkout.

The containment rule now has a single definition, `worktree.IsManaged`, shared by adoption, removal, rename and teardown. `filepath.Rel` over resolved paths, so `..`, a `.worktrees-evil` sibling, and a symlink escaping the directory all fail it.

**Fail-closed fixes** — three checks answered "safe to delete" when they had actually failed:

| Was | Now |
|---|---|
| `Inspect` treated *any* stat error as "missing, therefore pristine" | only ENOENT is disposable; EACCES/EIO report `Unknown` |
| `aheadCount` returned 0 on error, indistinguishable from "nothing unpushed" | reports whether it could answer; `Inspect` turns failure into `Unknown`. Also passes `--` so a ref that names a path isn't ambiguous |
| `projectRoot` used `Root`, not `MainRoot` | a project rooted in a linked worktree no longer puts `.worktrees/` in the wrong place and refuse every path |

**Also applied:** live-session re-check under `gitMu` in `RemoveWorktree`/`RenameWorktree` (both run off the read loop, so a session could start in the gap and be left in a deleted directory); dropped an unreachable `|| st.Unpushed > 0`; the delete dialog no longer promises commits are "recoverable" when uncommitted changes die either way; `DELETE_BRANCH` added to the `DESIGN.md` runOp list.

**Not applied:** `sanitizeBranch` doesn't reject `..`. Verified unexploitable — git's ref-name validation rejects it before any filesystem call on every path that reaches it. Containment is inherited from git rather than asserted locally, which is worth revisiting if a preview path is ever added that builds a worktree path without a git call in front of it.

**New regression tests:** the reproducer (project cwd on a linked worktree → close session → checkout intact), `Kill` refusing an unmanaged path, the browser working when the project cwd is a linked worktree, an unreadable worktree not reading as pristine, `IsManaged` edges including symlink escape, and the dialog copy in both jsdom and Playwright.

**Checks:** PASS — `go vet`, `go test ./...`, 517 unit/DOM, 172 mock e2e, 8 real-daemon e2e, `biome ci`, `tsc`.

**Threads:** 0 open before, 0 open after (the review returned findings in-session; nothing was posted as review threads).

PR #279, opened 2026-08-24
feat(worktree): detect squash merges, speed up the inventory

lucascaro:
## Why

The worktree browser's `merged` flag came from `--merged <default>` — reachability only. A squash-merged branch is not an ancestor of the default ref, so finished work read as unmerged: it sorted to the top of the orphan list as "may still hold work", and a squash-merged worktree showed *N commits ahead* and could only be deleted through the force path behind a red "this discards those commits" confirm.

On this repo (169 branches, 7 worktrees) that browser also took **~10.7s** to open, and a long branch list squeezed the worktree list out of the panel entirely.

## What changed

**Squash detection** — three stages, cheapest first:
1. reachability (`--merged`, unchanged)
2. patch id: the branch's cumulative diff (`base...branch`) matched against one shared index of the default ref's history. Replaces `git cherry` per branch.
3. `gh pr list --state merged` for squashes edited on the way in (conflict resolution, update-branch). Strictly additive: every failure mode — not installed, unauthenticated, offline, non-GitHub remote — yields nothing and can only leave branches looking *unmerged*. Cached 60s per repo; the only network call, and it never runs under the git lock.

Detection is a heuristic in the safe direction: an edited squash reads as unmerged, never the reverse.

**The verdict follows through** — a merged branch's unpushed commits stop being a delete blocker, in the client *and* in the daemon's own refusal (they must agree, or the UI offers a delete the daemon rejects). The local ref is removed with `-D`, since git's `-d` test is reachability too.

**Remote deletion** is now its own button rather than implied — a push is the one step that reaches beyond the machine — and only appears for a branch that tracks something. Orphan branches get `Delete local only` / `Delete local + remote`; worktrees get `Delete + local branch` / `Delete + branch everywhere`.

**Visibility**: `merged` is a badge on the row, not a word in the grey status line.

**Layout**: both panel sections are flex items, so they shrank *proportionally to content* — 120 branches against 3 worktrees left the worktree list 29px tall. Worktrees now hold the top half; the branch list takes what is left. Each scrolls on its own.

## Performance

Measured against a running daemon on this repo (169 branches, 7 worktrees):

| | before | after |
|---|---|---|
| `ListBranches` | 7.7s | 0.97s |
| per-worktree Inspect | 1.7s serial | ~0.4s parallel |
| **`ListWorktrees` end to end** | **~10.7s** | **1.43s warm / 1.56s cold** |

Four fixes: one `for-each-ref` with `%(ahead-behind:)` instead of 169 `rev-list` spawns; one shared patch-id index instead of `git cherry` re-walking history per branch; a bounded worker pool for the per-branch probes; and `gh` + branch listing + worktree probes now overlap instead of running back to back. The remaining ~1.4s is git-subprocess throughput, not a dependency chain.

## Verification

- `go test ./internal/...`, `go vet`, `-race` on the registry
- new Go tests: squash-merge detection (real repo with a `merge --squash`), squash-merged worktree is pristine, `gh` overlay reaches both lists + `RemoveWorktree` + `DeleteBranch` (stubbed — a `TestMain` keeps the whole package off the network)
- 524 vitest, 179 Playwright e2e (new: layout asserted in Chromium with `elementFromPoint`, since vitest is CSS-blind; remote-delete button shape)
- `tsc --noEmit`, `biome ci .`
- driven end to end against a live isolated daemon: 143/162 orphan branches and 5/7 worktrees correctly marked merged

## Notes for review

- `%(ahead-behind:)` needs git 2.41; older git falls back to the per-branch count.
- The patch-id index reads at most 2000 commits of the default ref — an older squash reads as unmerged.
- ~85 branches track something other than the default ref and still cost one `rev-list` each. Making "N ahead" mean *ahead of the default ref* for every branch would collapse that into the batched call (~1.4s → ~1.0s), but it changes the number on screen — deliberately left alone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   given a worktree. Deleting a branch that still holds unmerged commits
   says how many and requires a second confirmation.
 
+- The worktree browser now recognises squash merges. A branch merged
+  with GitHub's "Squash and merge" is not part of the default branch's
+  history — its commits were rewritten — so the browser used to report
+  finished work as unmerged, sort it to the top as "may still hold
+  work", and ask you to force-delete it past a warning about losing
+  commits. Such branches now carry a **merged** badge, sort down with
+  the rest of the finished work, and delete without the destructive
+  confirmation. Detection compares the branch's changes against the
+  default branch's history, and consults GitHub for the merges that
+  comparison cannot see (which needs the `gh` command; without it, or
+  offline, everything still works — some squashes just keep reading as
+  unmerged). A branch you have committed on since its merge is never
+  treated as merged, whatever GitHub says about it.
+
+- Deleting a branch from the worktree browser can now delete it on the
+  remote too. It is a separate button, never the default, and it only
+  appears for a branch that actually tracks a remote.
+
 - The new-session popup can now name the worktree's branch. Ticking
   "Create in git worktree" reveals a branch-name field; leave it empty to
   keep the previous behaviour of an auto-generated adjective-noun name.
@@ -85,6 +103,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   **Go Back** / **Go Forward**.
 
 ### Fixed
+
+- The worktree browser was slow to open on a repository with a lot of
+  branches — several seconds, during which the panel sat empty. It now
+  opens in about a second on a repository with 169 branches, by asking
+  git for every branch's state in one go instead of once per branch,
+  and by running the remaining checks at the same time as each other
+  rather than one after another.
+
+- A long list of branches used to squeeze the worktree list out of the
+  worktree browser: with a hundred branches the worktrees were a few
+  pixels tall and effectively unreachable. The panel is now a single
+  list — worktrees first, then the branches without one — with headings
+  that stay put as you scroll.
 
 - Reordering sessions sometimes moved the wrong one, or nothing at all.
   A session's order number is the position the daemon splices at, but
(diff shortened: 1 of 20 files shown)


lucascaro:
Review addressed in 036bf6b + 0943362.

**Blocking 1 — `gh` overlay keyed on branch name only.** Confirmed and fixed. The listing now requests `headRefName,headRefOid`, and a claim counts only when the local tip is reachable from the commit GitHub merged (`worktree.IsAncestor`). Anything unverifiable — OID never fetched, branch moved on, name reused — reads as unmerged.

Both regression cases are covered, and I checked they actually fail against the old behaviour rather than passing vacuously:

```
--- FAIL: TestRemoveWorktree_GHMergedDoesNotCoverLaterCommits
    got <nil>, want ErrWorktreeUnpushed
--- FAIL: TestDeleteBranch_GHMergedIgnoresUnrelatedHistory
    got <nil>, want ErrBranchUnmerged
```

The PR description's safety claim was wrong for the gh tier, exactly as you said; it now holds for both tiers.

**Blocking 2 — `Pristine()`'s blast radius.** Took the first option: `Pristine()` is back to ignoring `Merged`, with a comment naming both callers and why they must not widen. The browser's delete consults `Merged` on its own, where the act is explicit and confirmed. `TestInspect_DetectsSquashMergedWorktree` now asserts `Merged && !Pristine()` so a future re-widening trips.

**Important 3 — unbounded walk in `squashMerged`.** Capped with the same `--max-count=2000` as the index, so both paths answer from one window.

**Important 4 — untested destructive path.** `DeleteRemoteBranch` and `UpstreamOf` now have Go tests against a real bare remote, including the "already deleted" swallow.

**Minors:** 1 refused rather than ignored (`delete_remote` without `delete_branch` is now an error); 2 fixed (`refs/heads/`+branch); 3 fixed (`scrubURLCredentials` on push failures, with a test); 4 narrowed instead of cast. 5 kept deliberately — that timing line is what the perf work was measured with and it earns its keep.

Not taken: building the patch-id index once per `ListWorktrees` and passing it into `Inspect`. It is the better shape and I'd like it, but it changes `Inspect`'s signature for every caller including the two auto-delete paths, and that is a wider diff than this PR should carry. Worth its own change.

Also in this push, from live use rather than review: the browser's two panes became one scroll container (036bf6b) — as flex items they shrank proportionally to content, so a hundred branches left the worktree list a few pixels tall.

PR #281, opened 2026-08-24
fix(dev-iso): stop the stranded daemon on --reset, one iso dir per checkout

lucascaro:
Two bugs in `scripts/dev-iso.sh`, found the hard way: an iso run during #280 left four `claude` agent PTYs alive for half an hour after its window was closed, under a daemon belonging to a *different* worktree.

## 1. `--reset` wiped state but left the daemon running

`hived` keeps its registry in memory, so deleting the state dir does nothing to a daemon already listening on that socket. The GUI dials the survivor, adopts its in-memory sessions, and the clean slate you asked for silently didn't happen. Closing the window then leaves that daemon — and its agent PTYs — running for the next run to inherit.

`--reset` now stops the socket's owner first, reusing the pidfile `hived` already writes next to the socket for exactly this purpose (`cmd/hived/main.go:82`), with an `lsof` fallback when the pidfile is missing. Two safety properties:

- **Never signals a recycled pid** — the target's command line must still name that socket.
- **SIGTERM first**, escalating to SIGKILL only after 5s. `hived` shuts down cleanly on SIGTERM and removes its own pidfile.

## 2. Every worktree shared `/tmp/hive-iso`

Which is what allowed a daemon from another checkout to own this socket at all: all worktrees raced for one path, and whichever `hived` bound it first served everyone — with its own binary, its own `--cwd`, and its own sessions.

Default is now `/tmp/hive-iso-<checkout>`:

```
/Users/lucascaro/checkout/hive                        -> /tmp/hive-iso-hive
/Users/lucascaro/checkout/hive/.worktrees/fierce-cove -> /tmp/hive-iso-fierce-cove
/Users/lucascaro/checkout/hive/.worktrees/solid-creek -> /tmp/hive-iso-solid-creek
```

`--dir` still overrides, and the guards keeping `--reset` inside `/tmp` and `/var/folders` are untouched.

## Also

`--help` read a hard-coded `sed -n '2,12p'` line range, which this header change would have silently truncated. It now prints the comment block up to the first line of code.

## Verification

Exercised against a real `hived` (not a mock):

| Case | Result |
|---|---|
| Live daemon + pidfile | stopped via SIGTERM; pidfile self-removed |
| No dir at all | silent no-op, exit 0 |
| Stale pidfile naming a **live unrelated** process | innocent pid untouched |
| Stale pidfile naming a dead pid | silent no-op, exit 0 |
| Pidfile deleted, daemon still listening | found via `lsof`, stopped |

Plus all six destructive-path guards re-checked: `/tmp`, `/`, `$HOME`, `/tmp/../etc`, `/var/folders`, `/private/tmp` all still refuse with exit 2.

No CHANGELOG entry — dev-only tooling, matching how the file is used today.


lucascaro:
Added a second commit: the macOS leg was blocked by a pre-existing defect in `internal/agent/copilot_test.go`, unrelated to this PR (which changes one shell script and no Go).

```
panic: Fail in goroutine after TestCopilotCaptureIgnoresPreexistingSessionsInSameCwd has completed
  copilot_test.go:42   (os.Chtimes)
  copilot_test.go:130  (the goroutine)
  copilot_test.go:128  (created here)
```

The delayed writer goroutine was never joined, so it outlived the test. `t.TempDir()`'s cleanup then removed the directory between its `WriteFile` and its `Chtimes`; the `Chtimes` failed and called `t.Fatalf` from a goroutine after the test had returned. Go turns that into a panic that **replaces the real assertion message** — which is why the failing run never showed whether capture returned the wrong uuid or errored.

Fix: `writeCopilotSessionErr` reports through an error (the only form safe off the test goroutine), and `writeCopilotSessionAfter` spawns the delayed writer and registers the join in `t.Cleanup` — not at the end of the body, so it survives `t.Fatalf` exits, and cleanups being LIFO it runs before `TempDir`'s removal. Applied to both writer sites plus the partially-written-workspace test, whose goroutine swallowed errors but still wrote into a directory being deleted.

The race is narrow (`MkdirAll` recreates the tree, so only a cleanup landing between `WriteFile` and `Chtimes` trips it) — 300 `-race` runs pass locally, and it passed on `main` at `a7f0097`. The fix is structural rather than probabilistic: the writer cannot outlive the test, and never touches `t` off-thread.

PR #283, opened 2026-08-24
docs: drop the v1 TUI and the finished rewrite plan

lucascaro:
## Why

The tmux + Bubble Tea TUI (Hive v1) lives only on `release/v1` — no TUI code, dependency, CLI flag or build target exists on `main`. The GUI is the only option, so the v1/v2 branch banners were archaeology and "v2" is noise with no v1 to distinguish from.

## What

- **Deleted `docs/native-rewrite/`** (8 files, ~1360 lines) — the rewrite shipped. Its only unfinished items (notarization, installers, auto-update) moved into README's "Not yet shipping" line.
- **Stripped v1/TUI/tmux/`release/v1` framing** from README.md, AGENTS.md, DESIGN.md, golden-principles.md. Release tags (`v2.3.0`) kept — those are real.
- **Repointed dangling `phase-1.md` refs** in `cmd/hived/main.go`, `cmd/hivegui/main.go`, `cmd/hivegui/README.md` at `DESIGN.md`; dropped "Phase 1" from `internal/daemon/socket.go`.
- **Rewrote SECURITY.md** — it was stale beyond tmux: scoped a `hive` binary and `hive mux-daemon` that don't exist, named the wrong socket path (`~/.config/hive/mux.sock`) and log files, and documented a hooks feature that shipped only in v1. Now written against `internal/daemon/socket.go` (dir mode `0o700`) and `internal/registry/logfile.go` (logs `0o600`).

## Out of scope, deliberately

- `docs/exec-plans/`, `docs/product-specs/`, `docs/analysis/`, `CHANGELOG.md` — frozen records. Their `phase-1.md` links now dangle by design.
- Code/test comments where "TUI" means a guest full-screen app inside a session (vim, htop, claude) — correct usage.

## Verification

- `grep -rniE '\btui\b|bubble ?tea|tmux|release/v1'` over the living docs → zero hits.
- No `native-rewrite` refs outside the frozen historical dirs.
- `go build ./internal/... ./cmd/hived/...` and `go vet` pass. (`./cmd/hivegui` needs a built `frontend/dist`, absent in this worktree — pre-existing, unrelated.)


lucascaro (commented):
## SCOPE

17 files · ~1620 changed lines (1360 of them pure `docs/native-rewrite/` deletion, 0 lines of behavior change) · base: `main`

Investigated: **deletions & renames** (the only triggered angle — `docs/native-rewrite/` removed, `phase-*.md`/`PLAN.md` symbols grepped repo-wide) plus factual verification of the rewritten `SECURITY.md` against `internal/daemon/socket.go` and `internal/registry/logfile.go`.
Skipped: call sites (no signature changed), implementations (no interface touched), consumers (no schema/wire change), supply chain (no dependency touched), convention drift (no new helper), test reach (Go changes are comments only).

Dimensions checked: **correctness** (Go diff is three comment edits; no logic reached) · **safety/test hygiene** (no tests, no golden files, no module-level state — clean) · **security** (SECURITY.md claims verified line-by-line against source; see IMPORTANT 3) · **performance/UX/consistency** (where the findings are).

Size gate note: the diff nominally exceeds the split threshold, but 84% of it is one directory of deleted prose and nothing executable changed, so Pass 1 ran inline — four dimension agents would have re-read the same deleted markdown four times.

The core claim of the PR holds up. `hive mux-daemon`, `~/.config/hive/mux.sock`, `mux-daemon.log` and `~/.config/hive/hooks/` genuinely do not exist on `main` — the old SECURITY.md was documenting a program that isn't shipped, which is worse than documenting nothing. Every new factual claim I could check is correct: socket paths match `SocketPath()` for darwin/linux/windows, `EnsureSocketDir` really does `MkdirAll(dir, 0o700)`, `hived.log`/`hivegui.log` are the real names (`cmd/hived/main.go:55`, `cmd/hivegui/main.go:55`) and `OpenLogFile` really opens them `0o600`. README's "not yet shipping: code signing and notarization, platform installers, auto-update" is also accurate — no `codesign`, `notarytool`, or updater code exists anywhere in the tree.

## BLOCKING (must fix before merge)

None.

## IMPORTANT (should fix, could be fast follow-up)

**1. [cmd/hivegui/README.md:63] (confidence: 9/10) Half-updated file: dangling `PLAN.md` link to a file this PR deletes, plus four false "Phase N" claims**

Why: The PR touched this file — it dropped "for Phase 1" from line 3 and repointed line 8 at `DESIGN.md` — then stopped. Line 63 still says *"see PLAN.md Phase 2+"*, and `docs/native-rewrite/PLAN.md` is deleted in this same diff. It is now the **only** remaining `PLAN.md` reference in the repository, so this is a self-inflicted broken link, not one of the "frozen record" danglers the PR body deliberately scopes out (`cmd/hivegui/` is live docs, not `docs/exec-plans/`). Worse, lines 45–95 are actively wrong about today's product: "Phase 1 acceptance test", "**One session, one window. Multi-session is Phase 2**", "No sidebar, no naming, no reordering", "No agent integration. Phase 3" — all four contradict `README.md` and `DESIGN.md`, which describe a shipping multi-session sidebar with names, colors and an agent catalog at v2.3.0. `AGENTS.md:150` ("stale docs are a bug") and `AGENTS.md:173` ("if a doc becomes incorrect after your change, fix it in the same commit") both apply, as does the boil-the-lake stance at `AGENTS.md:281`.

Fix: complete the file in this PR rather than leaving it for `/hs-doc-garden`:
- L45 → `## Acceptance test (manual)`
- L63–64 → drop the `PLAN.md` sentence; daemon-restart resume is already tracked in `README.md`'s "Not yet shipping" line, so just say "you get a fresh shell — daemon-restart resume is not implemented."
- L85 → drop "Phase 2 may suppress cursor flicker during replay" or restate it as a plain TODO with no phase number.
- L90–95 → replace the whole "Known limitations (Phase 1)" block with what is actually true of `hivegui` today; if nothing on that list still holds, delete the section outright.

**2. [docs/design-docs/index.md:9] (confidence: 8/10) Deleting `phase-0-report.md` drops the only record of three still-load-bearing decisions, with no successor doc**

Why: This is the one place where the deletion loses something git history alone doesn't replace. `phase-0-report.md` was an ADR in all but name, and three of its decisions still govern code on `main`:
- **Why `aymanbagabas/go-pty` and not `creack/pty`** — "`creack/pty` is POSIX-only… go-pty exposes a single `Pty` interface backed by Unix PTYs on POSIX and ConPTY on Windows." A future agent proposing to "simplify" back to `creack/pty` has nothing on `main` telling it why that breaks Windows.
- **Why AF_UNIX on Windows instead of named pipes** — the "Windows 10 1803+" constraint is asserted verbatim in the comment this very PR edits (`internal/daemon/socket.go:11`) and its justification now exists nowhere in the tree.
- **The kitty-keyboard passthrough finding** — why a passthrough client must own terminal emulation, and why the GUI route made the problem disappear. That is exactly the kind of thing someone re-derives painfully.

`AGENTS.md:172` says "record non-obvious architectural decisions under `docs/design-docs/`", and that directory's Active list (this line) is currently empty. Deleting the de-facto ADR without migrating it leaves the documented home for decisions empty while the decisions themselves stay in force. Everything else in the directory — the phase plans, milestone tables, manual acceptance scripts, the superseded wire-protocol v0 sketch — is genuinely finished work and correct to delete.

Fix: before merging, distill `phase-0-report.md` into one `docs/design-docs/native-rewrite-decisions.md` (roughly a page: Wails v2 + xterm.js, go-pty over creack/pty, AF_UNIX over named pipes with the 1803 floor, and the kitty-keyboard lesson — each with its constraint and the alternative rejected), add the row to `index.md:9`, and point `internal/daemon/socket.go:11` at it. That converts the deletion from lossy to lossless.

**3. [internal/daemon/socket.go:46] (confidence: 7/10) SECURITY.md now promises a socket guarantee the code does not enforce**

Why: `SECURITY.md:34-35` states "the containing directory is created with mode `0o700`, so only the owner can reach the socket", and `SECURITY.md:41-42` repeats it under **Multi-User Systems** — the exact section where an attacker is in the threat model. But `os.MkdirAll(dir, 0o700)` applies the mode **only when it creates the directory**; on a pre-existing directory it returns `nil` without touching permissions or verifying ownership. On macOS, and on Linux whenever `XDG_RUNTIME_DIR` is unset, that directory is `/tmp/hive-<uid>` — inside world-writable `/tmp`. On a shared box another user can pre-create `/tmp/hive-1000` mode `0777` before the victim's first `hived` start; `MkdirAll` accepts it, and the dir's owner can then unlink the bound socket and bind their own, MITM'ing every agent session. The prior wording claimed a `0o600` mode on the socket *file*, which was at least a property of the socket itself. The code weakness predates this PR — but a security policy is exactly the document that shouldn't assert an unverified guarantee, and the PR body explicitly presents this file as "now written against `internal/daemon/socket.go`".

Fix: make the code match the doc, since it's four lines:
```go
func EnsureSocketDir(sockPath string) error {
	dir := filepath.Dir(sockPath)
	if err := os.MkdirAll(dir, 0o700); err != nil {
		return err
	}
	if runtime.GOOS == "windows" {
		return nil
	}
	fi, err := os.Stat(dir)
	if err != nil {
		return err
	}
	if st, ok := fi.Sys().(*syscall.Stat_t); ok && int(st.Uid) != os.Getuid() {
		return fmt.Errorf("socket dir %s is owned by uid %d, not %d", dir, st.Uid, os.Getuid())
	}
	if fi.Mode().Perm() != 0o700 {
		return os.Chmod(dir, 0o700) // pre-existing dir: MkdirAll left the mode alone
	}
	return nil
}
```
If you'd rather not touch code in a docs PR, soften `SECURITY.md:35` and `:41` to "the containing directory is created with mode `0o700` when Hive creates it; on a shared machine verify `/tmp/hive-<uid>` is owned by you and mode `0700`" — and open a follow-up issue for the `Chmod`/ownership check. Either is fine; asserting the guarantee with neither is not.

**4. [cmd/hived/main.go:1] (confidence: 8/10) Package comment still says the daemon owns *one* session — contradicting the DESIGN.md it now points at**

Why: The PR rewrote line 3 of this comment to read "See `DESIGN.md` for the role of this binary in the architecture", but left lines 1–2 saying `hived` "owns **one** PTY-backed shell session". `DESIGN.md:14` describes the daemon as a "**multi-session** PTY host", and the registry holds many sessions. The new pointer now sits directly beneath a sentence the target document contradicts — a v1-era leftover of exactly the kind this PR set out to remove, in a line it edited.

Fix: `// hived is the Hive session daemon. It hosts PTY-backed shell sessions and accepts client connections over a Unix socket. See DESIGN.md for the role of this binary in the architecture.`

**5. [DESIGN.md:51] (confidence: 7/10) "GUI restart preserves scrollback" loses its last normative home**

Why: `docs/product-specs/143-...md:13` and `docs/exec-plans/completed/143-...md:23` both cite `docs/native-rewrite/phase-1.md:157` and `phase-2.md:14-16,212-216` as the *normative source* for the contract that a GUI restart preserves scrollback — that citation is what made PR #141 a regression rather than a design change. I confirmed the deleted files carry it (phase-1 milestone 1.8 "Acceptance: vim survives GUI restart"; phase-2's pass criteria "clicking each session attaches and shows scrollback"). That invariant is still live and distinct from README's "not yet shipping: scrollback resume across **daemon** restart" — but it appears in none of DESIGN.md's Hard rules. Treating the 143 docs as frozen records is right; letting the invariant they cite evaporate is not.

Fix: add one Hard rule under this section — **"GUI restart preserves scrollback."** A `hivegui` restart against a live `hived` replays prior session content; only a *daemon* restart may start clean. Then the frozen 143 docs' dangling links become harmless history rather than the last trace of a live contract.

## MINOR (nice to have)

1. [SECURITY.md:33] Linux path is incomplete — `SocketPath()` falls back to `/tmp/hive-<uid>/hived.sock` when `XDG_RUNTIME_DIR` is unset (and the same branch serves freebsd/openbsd/netbsd), which is the case where the /tmp exposure in IMPORTANT 3 actually bites.
2. [SECURITY.md:31] Neither `HIVE_SOCKET` nor `HIVE_STATE_DIR` is mentioned, though both relocate the socket and the `0o600` logs out of every path this section documents.
3. [SECURITY.md:47] The removed "Hook Scripts" section had no successor for one behavior that survived v1: `internal/worktree/worktree.go:230` symlinks `.claude/hooks` (along with `plugins`, `commands`, `skills`) into every new worktree, so Hive actively propagates user-supplied executable config into fresh checkouts. "Hive does not sandbox them" covers the agent, not Hive's own act of linking. One sentence in **Agent Processes** would close it.

## Verdict

**COMMENT**

Correct and well-motivated cleanup — the old SECURITY.md documented a binary that doesn't exist, and every new claim in the rewrite checks out against the source — but it stops short in three places: `cmd/hivegui/README.md` is left half-updated with a dangling link to a file this PR deletes, `phase-0-report.md`'s still-binding decisions have no successor in the directory `AGENTS.md` designates for them, and `SECURITY.md` asserts a socket guarantee `MkdirAll` doesn't actually enforce.

PR #286, opened 2026-08-28
Improvement pass: un-quarantine e2e-real, Go static analysis, god-file splits

lucascaro:
Implements the improvement plan in `docs/analysis/2026-08-24-improvements.md`
(written at the start of this branch; its status table records what landed,
what did not, and why).

## The headline: the "flaky" e2e-real suite was never flaky

All three specs quarantined under spec 245 failed **deterministically** —
0/10 local runs, same three tests — because each guarded against vacuity
with `traceTags(page, 'replay-request') > 0` in a scenario where the tile is
*following the bottom*. `decideResizeReplay` deliberately SKIPS the
destructive full-ring replay for a follower; that skip is itself a shipped
fix for the renderer freeze under live output. So a healthy follower emits
`replay-skip` and exactly zero `replay-request`, and the guard failed against
correct code, every time.

Meanwhile CI ran **2 of 12** e2e-real tests and reported green for weeks.

The guard now counts resizes that reached the replay *decision* (request +
skip) instead of demanding one outcome. Every invariant the specs exist for is
untouched: markers exactly once and in order, viewport converges to the bottom,
a follower is never stranded mid-history.

| | before | after |
|---|---|---|
| the 3 specs, `--retries=0`, macOS idle | 0/10 runs green | 10/10 runs green |
| full `e2e-real`, `CI=true` | 11 passed / 10 skipped | 21 passed, 3/3 runs |

Quarantine removed. `retries: 1` stays, now paired with `failOnFlakyTests`, so
a retry buys the diagnostics of a second attempt — not a green check. Green
means first-attempt green, which is spec 245's own success criterion made
structural instead of a thing someone has to remember to audit.

## Two real defects found along the way

- **`TestE2E_DaemonRestart` raced the async revive.** It asserted `alive:true`
  on the first snapshot after a restart; since #282 the daemon binds the socket
  before `reviveAll` forks any PTY and parks restored entries in `reviving`, so
  that snapshot is legitimately `alive:false`. Failed 3/3 locally, passed CI on
  timing luck. Now polls until the entry comes up.
- **`hived-ws-bridge` accepted an off-host bind.** `CheckOrigin` returned true
  for everything on the strength of a comment saying the listener was
  localhost-only, while `-addr` would happily take `0.0.0.0`. `requireLoopback`
  now enforces what the comment claimed — the bridge speaks for a real daemon
  with no auth of its own.

## Everything else

- **Go static analysis in CI**: `staticcheck` (pinned 2025.1.1) + `govulncheck`
  on the Linux leg. Baseline was one real finding `go vet` misses.
- **God-files**: `app.go` 1101 → 192 + four domain files. `serveControl` 274 →
  144, dispatch lifted to `handleControlFrame(...) bool` behind a `controlOps`
  value; twelve identical decode prologues collapsed into `decodeReq[T]`.
  `registry.kill` 164 → 128 by extracting `disposeWorktree`.
- **Coverage where the seams opened**: `internal/daemon` 73.0 → 78.4%,
  `cmd/hivegui` 30.5 → 38.2%, `hived-ws-bridge` 30.7 → 39.8%. Includes the
  first test that registry teardown actually consults `IsManaged` — the guard
  that stops hive deleting a directory it does not own.
- **Deps**: Go module graph refreshed (`go-pty` 0.2.3 + indirects).
- **Plan-lifecycle drift**: 247 → `completed/`; 142's stub corrected (PR #160
  was closed unmerged, the bug is still real); `scripts/check-plan-lifecycle.sh`
  so it stops recurring.
- Stray root `node_modules/` removed (a vitest cache committed by accident in
  #255) and gitignored.

## Deliberately not done

- **xterm 6.0** — compiles fine (none of the removed APIs are used), but 6.0
  rewrote the viewport/scroll implementation and `src/` has 87 call sites on
  exactly that surface, in the subsystem with this repo's longest bug history.
  Needs a full 5000-line buffer in a real WebView, which no CI leg runs.
- **Wails 2.15** — `ci-bootstrap.sh` pins the *CLI* to 2.12.0 and the CLI
  generates the bindings; library-only bump invites skew.
- **Splitting `session-term.ts` / `style.css`** — pure file-moving, no seam
  unlocked, unlike the two splits here that both bought testability.

One latent hazard is logged to the tech-debt tracker rather than patched:
`disposeWorktree`'s non-git fallback does `os.RemoveAll(wtPath)` with the
`IsManaged` guard skipped, contradicting the branch right below it. Fixing it
either way is a behaviour change.

## Verification

Local, all green: build, vet, staticcheck, govulncheck, full Go suite, Go e2e
(`-tags=e2e`), tsc, Biome, spec discovery, 530 unit tests, 182 mock e2e, 21
e2e-real.

<h3>Greptile Summary</h3>

The PR broadens CI and end-to-end coverage, adds static analysis, fixes daemon-restart test synchronization, splits large Go files, and hardens the test WebSocket bridge.
- Restores the previously quarantined real-daemon end-to-end tests with corrected replay-decision assertions.
- Adds staticcheck and govulncheck to the Linux CI leg.
- Refactors GUI control, attach, spawn, and RPC logic into focused files with added tests.
- Completes the loopback-listener fix by accepting only literal loopback addresses before binding.

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge.

No blocking failure remains.

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| cmd/hived-ws-bridge/main.go | The revised bind guard rejects hostnames and accepts only loopback IP literals, closing the previously reported resolution-to-bind gap. |
| cmd/hived-ws-bridge/main_test.go | Adds focused coverage for loopback bind validation, origin checks, isolation checks, and concurrent frame writes. |
| .github/workflows/ci.yml | Adds pinned Go analysis tools and expands real-daemon end-to-end execution across supported runners. |
| cmd/hivegui/app_control.go | Extracts control-frame dispatch from the former god file into a testable control-operations seam. |
| internal/registry/registry.go | Extracts worktree disposal while retaining the existing ownership and cleanup behavior. |

</details>

<sub>Reviews (12): Last reviewed commit: ["test: quarantine the second e2e-real tes..."](https://github.com/lucascaro/hive/commit/0789e4e12c9106e16995fd86868b5defdd2dd26a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57789433)</sub>


cmd/hived-ws-bridge/main.go
```diff
@@ -84,10 +87,40 @@ func requireIsolation() error {
...
+	host, _, err := net.SplitHostPort(addr)
+	if err != nil {
+		return fmt.Errorf("addr %q: %w", addr, err)
+	}
+	if host == "" {
+		return fmt.Errorf("addr %q binds every interface; use 127.0.0.1 or [::1]", addr)
+	}
+	if host == "localhost" {
+		return nil
+	}
```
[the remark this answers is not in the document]
lucascaro: Valid, and fixed in 7420892 — thank you.

Both comments are the same defect seen twice: `requireLoopback` resolved the name, checked the addresses, then discarded them, while `net.Listen` re-resolves the name independently. That is check-then-use, and on a bridge fronting unauthenticated daemon operations (PTY stdin, worktree removal) it is worth closing properly rather than narrowing.

So instead of binding the resolved IP, names are now refused outright: the guard's answer is about the exact literal `net.Listen` will use. Nothing needed a name — the only caller, e2e-real's `globalSetup`, passes `127.0.0.1:0` — so the fix deletes the `LookupIP` path rather than adding to it.

`localhost` remains acceptable as an **Origin**: that check reads a header on an already-established connection, involves no resolution or binding, and the harness's Vite dev server does send it.

Verified: vet + staticcheck clean on darwin/linux/windows, the full Go suite, and e2e-real 22/22 — which is what proves the bridge still binds and the harness still connects.

cmd/hived-ws-bridge/main.go
```diff
@@ -84,10 +88,74 @@ func requireIsolation() error {
...
+		}
+		return nil
+	}
+	if host != "localhost" {
+		return fmt.Errorf("addr %q: host is not an IP literal or localhost", addr)
+	}
+	// Resolve rather than trust the name: a hosts file can map
+	// localhost anywhere, and "it said localhost" is not the property
+	// this guard is asserting.
+	ips, err := net.LookupIP(host)
```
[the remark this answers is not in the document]
lucascaro: Valid, and fixed in 7420892 — thank you.

Both comments are the same defect seen twice: `requireLoopback` resolved the name, checked the addresses, then discarded them, while `net.Listen` re-resolves the name independently. That is check-then-use, and on a bridge fronting unauthenticated daemon operations (PTY stdin, worktree removal) it is worth closing properly rather than narrowing.

So instead of binding the resolved IP, names are now refused outright: the guard's answer is about the exact literal `net.Listen` will use. Nothing needed a name — the only caller, e2e-real's `globalSetup`, passes `127.0.0.1:0` — so the fix deletes the `LookupIP` path rather than adding to it.

`localhost` remains acceptable as an **Origin**: that check reads a header on an already-established connection, involves no resolution or binding, and the harness's Vite dev server does send it.

Verified: vet + staticcheck clean on darwin/linux/windows, the full Go suite, and e2e-real 22/22 — which is what proves the bridge still binds and the harness still connects.

.github/workflows/ci.yml
```diff
@@ -91,6 +91,48 @@ jobs:
       - name: Vet
         run: go vet ./...
 
+      - name: Cache Go analysis tools
```
lucascaro, line 94: **IMPORTANT** (confidence 8/10) — this cache buys nothing as written.

It restores `~/go/bin`, and then both steps below run `go install ...@version` unconditionally, which always recompiles and overwrites the restored binary. So the stated purpose ("without this they are rebuilt on every run") is not achieved — what's left is cache restore/save latency on the gate every PR waits on, with no rebuild avoided.

Complete fix — guard both installs on the cached binary:

```yaml
# Staticcheck (Go)
run: |
  BIN="$(go env GOPATH)/bin/staticcheck"
  [ -x "$BIN" ] || go install honnef.co/go/tools/cmd/staticcheck@2025.1.1
  "$BIN" ./...
```

```yaml
# Vulnerability scan (govulncheck)
run: |
  BIN="$(go env GOPATH)/bin/govulncheck"
  [ -x "$BIN" ] || go install golang.org/x/vuln/cmd/govulncheck@v1.7.0
  "$BIN" ./...
```

The cache key already pins both versions, so a bump still misses exactly once — that part is right.

cmd/hivegui/frontend/test/e2e-real/bridge-sessions.ts
```diff
@@ -0,0 +1,143 @@
...
+  // the hand-written ws-shim.d.ts (see there for why not @types/ws).
+  const WS = globalThis.WebSocket ?? (await import('ws')).WebSocket;
+  const ws = new WS(WS_URL);
+  await new Promise((res, rej) => {
+    ws.onopen = res;
+    ws.onerror = rej;
+  });
+  const send = (id: number, method: string, params: object = {}) =>
+    ws.send(JSON.stringify({ id, method, params }));
+  const waitFor = (id: number) =>
```
lucascaro: **IMPORTANT** (confidence 6/10) — unbounded await in the new shared helper.

`waitFor(id)` here and `reply(id)` at :90 never reject, and `onerror` is only wired *before* open. A socket error after the handshake, or a bridge that simply never answers, hangs until Playwright's 90s test timeout and surfaces as "test timeout exceeded" with no cause attached.

`killAndAwaitRemoval` already has a 15s race guard on the removal wait — the request paths don't. In a PR whose whole point is making e2e failures legible instead of "flaky", this is the last place a failure arrives unlabelled.

Complete fix — apply to both `bridgeCalls` and `killAndAwaitRemoval`:

```ts
const waitFor = (id: number, ms = 10000) =>
  new Promise<{ id: number; error?: string }>((res, rej) => {
    const t = setTimeout(() => rej(new Error(`bridge: no reply to id ${id}`)), ms);
    ws.addEventListener('message', function h(ev) {
      const m = JSON.parse(ev.data);
      if (m.id === id) {
        clearTimeout(t);
        ws.removeEventListener('message', h);
        res(m);
      }
    });
    ws.addEventListener('error', (e) => { clearTimeout(t); rej(e); });
  });
```

cmd/hivegui/frontend/test/e2e-real/grid-focus-scroll.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+  }, WS_URL);
+});
+
+async function focusFirstTerm(page: Page) {
+  await page.evaluate(() => {
+    const helper = document.querySelector<HTMLTextAreaElement>(
+      '.term-host .xterm-helper-textarea',
+    );
+    if (!helper) throw new Error('no xterm helper textarea');
+    helper.focus();
```
lucascaro: **MINOR** — this regression guard's own helper does the exact thing the fix removes: `helper.focus()` with no `preventScroll`.

Harmless where it currently sits (both calls happen before the flood, so `.term-body` has no scroll slack yet), but it means the spec's fidelity depends on call ordering rather than on the fix. Moving or adding a `focusFirstTerm()` after the flood would make the guard fail against *correct* code — which is precisely the failure mode spec 245 was about.

Worth a one-line comment here saying why the bare `focus()` is deliberate, or just pass `{ preventScroll: true }` and rely on the app's own focus path for the assertion.

cmd/hivegui/frontend/src/app/focus.ts
```diff
@@ -124,11 +124,24 @@ document.addEventListener(
...
+// is `overflow: hidden` and — because the helper textarea sits below the
+// visible rows — has roughly a viewport's worth of scroll slack. The
+// browser therefore scrolls the terminal completely out of its own box
+// and the tile renders solid black, until any resize re-lays-out and
+// clamps scrollTop back to 0. That is the whole of the "black tile until
+// I resize" bug: the content was always painted, just scrolled out of
+// sight.
+// (xterm's own Terminal.focus() already passes preventScroll, so these
+// two call sites were the only ones that could trigger it.)
+const FOCUS_OPTS: FocusOptions = { preventScroll: true };
```
lucascaro: **MINOR** — `FOCUS_OPTS` is declared after its first lexical use at :127. Legal, since the use is inside a listener that runs long after module evaluation, but it reads as a TDZ hazard on first pass. Moving the `const` above the `focusout` listener costs nothing and keeps the comment attached to it.

lucascaro (commented):
## Review — /hs-review-pr

**Scope:** 56 files · ~4800 changed lines · base `main`.
**Investigated:** call sites (notify/daemon/registry symbol changes), deletions (the 909-line `app.go` split), supply chain (go-pty 0.2.2→0.2.3, 13 indirect bumps, `go` directive 1.25.0→1.25.14), convention drift (`bridgeCalls` extraction), test reach.
**Skipped:** implementations (no interface touched), consumers of changed data (no wire/schema change).
**Verified locally:** `go build ./...`, `go vet ./...`, `go test ./...`, `biome ci .`, `tsc --noEmit` — all green.

Substance is solid. Three things I checked rather than assumed:

- The `app.go` split and the `daemon.go` `handleControlFrame` extraction are **verifiable pure moves** — the exported+unexported function set and the whitespace/comment-normalized bodies both diff clean against the merge-base. `continue` → `return false` preserves loop semantics.
- `requireLoopback` refusing names outright (rather than resolving one and letting `net.Listen` resolve another) is the right call and closes the check-then-use gap the open Greptile thread describes.
- The `notify` platform split is the correct fix for U1000: the callback was written in shared code and read only from `notify_darwin.go`.

No blocking findings. Two worth fixing, both inline.

### Also
- The branch is currently `CONFLICTING` with `main`.
- The Greptile thread on `main.go:117` is still open upstream; the code it flags is already fixed here, so it just needs resolving.

**Verdict: COMMENT** — no blockers, two IMPORTANT items that can land here or as a fast follow-up.

.github/workflows/ci.yml
lucascaro: Fixed in ea0e572 — and you are right that the comment made it worse by claiming a benefit it did not deliver. Both steps now guard the install on the cached binary exactly as you wrote, and the comment now says a cache hit avoids the rebuild *because* of that guard. Key still pins both versions, so a bump misses once.

cmd/hivegui/frontend/test/e2e-real/bridge-sessions.ts
lucascaro: Fixed in ea0e572. Both `bridgeCalls` and `killAndAwaitRemoval` now share one `awaitReply(id, ms = 10000)` that rejects naming the request it was waiting on, and registers the `error` listener for the whole wait rather than only around open. Your framing is the part that decided it: in a PR about making e2e failures legible, an unlabelled 90s timeout was the last one left.

cmd/hivegui/frontend/test/e2e-real/grid-focus-scroll.spec.ts
lucascaro: Took the second option in ea0e572 — the helper passes `{ preventScroll: true }` and the assertion rests on the app's own focus path during ⌘G / ⌘2 / ⌘1. Re-ran red/green afterwards to be sure it did not weaken the guard: still red at `scrollTop=352` without the fix, green with it.

cmd/hivegui/frontend/src/app/focus.ts
lucascaro: Moved in ea0e572 — `FOCUS_OPTS` is now at :117, above both uses (:140 and :214), with its comment attached.

PR #338, opened 2026-09-04
feat(state): session state model — daemon-owned state, Claude hook tier (336)

lucascaro:
## Summary

Spec 336, phases 1–2: every session carries a daemon-owned `state` (`working` / `idle` / `waiting_input` / `waiting_permission` / `exited` / `error`) plus `state_source`, `last_prompt`, `last_summary`. `needs_attention` is derived from state; nothing else owns it.

- **Heuristic tier** (every session): screen-digest sampling, quiet timeout, bell → `waiting_input`, keystroke or switch-to clears.
- **Hook tier** (Claude): `hived hook` subcommand wired through `--settings` at spawn/resume/restart; `ModeEvent` on the socket; `HIVE_SESSION_ID` / `HIVE_SOCKET` in every child env. `DaemonContract` 2 → 4.
- **GUI**: local attention set deleted; glyphs and pulse render from the session list. User-input signal is `onKey`, not `onData`.
- **Evidence**: `HIVE_DEBUG_STATE=1` logs every transition with its cause; the transition table is frozen in the exec plan.

Bugs found and fixed along the way, each with a test that fails without the fix:
- terminal query replies (cursor position, focus reports) cleared the bell within a frame — fish, pi and Claude all provoke them
- the bell was ignored once hooks owned a session, so a finished Claude turn raised nothing
- `waiting_permission` could stick: Claude fires no hook between allow and tool end, and a declined question tool fires nothing at all

## Verification

- `go build/vet` (darwin/linux/windows), `go test -race ./internal/... ./cmd/hived/...`, staticcheck, `scripts/check-daemon-contract.sh`
- Frontend: `biome ci`, typecheck, vitest (1021), Playwright mock (270), Playwright real-daemon (26, incl. new `state-bell.spec.ts` for checklist rows 1–5)
- Recorded Claude PTY fixtures replayed in CI (`internal/session/state_fixture_test.go`)
- Live probes against claude 2.1.261 (`HIVE_PROBE_CLAUDE=1 go test ./cmd/hived/ -run TestClaudeProbe -v`): prompt → working, permission prompt → waiting_permission, allow → working, Stop → idle, idle_prompt → waiting_input, /exit → exited

Not in this PR: Pi extension (phase 3), hivebar count and notification wording (phase 4).

Plan: docs/exec-plans/active/336-session-state-model.md


lucascaro:
## Review loop escalated — 2 iterations, 5 items need a human decision

**Status:** no BLOCKING findings; all 13 checks green on `ac893ca`; mergeable; zero unresolved review threads.

Iteration 2 applied 6 safe fixes in one batch commit (`ac893ca`). The loop stopped because autofix surfaced five items it will not auto-apply — each is a design or taste call, not a mechanical edit.

### Needs a decision

1. **`internal/registry/registry.go:409` — out-of-order hook events.** Events are applied in arrival order; concurrent connections can invert them. Two valid designs: a per-machine `lastEventAt` guard, or per-session serialization in the daemon. Concurrency-semantics choice.
2. **`cmd/hivegui/frontend/src/app/events.ts:275` — a dead session is unreachable by ⌘B** and its pulse gets wiped. Fix needs a new `deadUnseen` Set threaded through three readers across three files.
3. **`cmd/hivegui/frontend/src/lib/session-state.ts:81` — a live session with an agent-reported error renders "Exited with an error."** Fixing it needs either a new state + icon + sprite + doc row, or a semantic reorder. Both change UI vocabulary.
4. **`cmd/hived/claude_probe_test.go` — the spawned real `claude` inherits the developer's `HOME`** and writes to the real `~/.claude`. The env-restore half was applied; the config-dir redirect was not, because the right env var could not be verified.
5. **`internal/session/state_fixture_test.go:288` — `TestFixturePermissionSkipped` is an unconditional `t.Skip` with no assertions.** Reclassified SAFE → RISKY: `testdata/state/README.md:69` cross-references it by name, so removing it is a coupled two-file edit plus a call on whether a documented-omission test is legitimate.

### One thing to eyeball in the applied fixes

The new `Apply` exited-guard returns `false` after setting `m.source` / `m.hookSeenAt`, so a source-only change on a dead session is not broadcast. Harmless (a dead session's tier does not render), but it is a deliberate asymmetry rather than an oversight.

### Suggested next step

None of the five block merge. Either pick a design for each and re-run `/hs-review-loop 338`, or ship as-is and open follow-ups — then run `/hs-merge-gate 336`.

PR #339, opened 2026-09-05
fix(daemon): don't unlink a replacement daemon's live socket

lucascaro:
`Daemon.removeOwnSocket` identified "is this still my socket?" with `os.SameFile` (dev+inode). Inode numbers get reused, so a replacement daemon's brand-new socket at the same path can compare equal to the one the dying daemon bound — measured on Linux `/tmp`: same `ino=2018165`, mtimes 1 ms apart, `SameFile == true`. The old daemon then unlinks the live daemon's socket, leaving it serving an inode nobody can dial. That is the GUI "Restart Daemon" path, and it is why `TestCloseLeavesAReplacementSocketAlone` was flaky on Linux CI (~1 full-package `-race` run in 4 at 2 CPUs; reproduced in a `golang:1.25` container against unmodified `main`).

## Fix

After `d.ln.Close()`, dial the path. Our own listener is gone, so anything that still accepts a connection is another daemon's. `SameFile` stays as the cheap "definitely not mine" gate — only its false positives needed catching.

## Tests

- `TestRemoveOwnSocketLeavesALiveSocketAlone` fabricates the inode collision exactly (a `Daemon` whose `sockInfo` is the stat of a live daemon's socket) and asserts the file survives and still dials. Verified red without the guard.
- `TestRemoveOwnSocketRemovesADeadSocket` keeps the other half honest.
- 18 × `go test -race ./internal/daemon/ -count=3` on Linux/2 CPUs: clean, against a ~1-in-4 baseline failure rate.

### DIFF
--- a/.changesets/daemon-socket-inode-reuse.md
+++ b/.changesets/daemon-socket-inode-reuse.md
@@ -1,0 +1,13 @@
+---
+issue: null
+pr: null
+type: fixed
+bump: patch
+---
+- Restarting Hive no longer risks leaving the daemon unreachable. When
+  the old `hived` shut down while its replacement had already taken
+  over the socket path, it could delete the replacement's socket —
+  inode numbers get reused, so "is this still my socket?" answered yes
+  for a file the old daemon never bound. The dying daemon now asks the
+  socket instead of the filesystem, and leaves any socket that is still
+  being served alone.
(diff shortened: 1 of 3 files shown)


lucascaro:
Applied `daemon-contract-override`: this changes only teardown behaviour inside the dying daemon (which socket file it unlinks). No wire frames, fields, or client-visible semantics change, so a GUI built from this tree drives an older daemon exactly as before — and a needless bump would cost every user their running sessions.

PR #347, opened 2026-09-05
fix(gui): right-align session count on minimized project chips

lucascaro:
Session count on minimized project chips is now right-aligned, with the attention badge immediately to its left, so counts line up down the sidebar list.

- `Chip.tsx`: alert renders before count.
- `minimized.css`: `.hv-chip__label { flex: 1 }` inside `#minimized-projects` so the label eats the slack.

**Tradeoff:** the old comment cited spec 255 — the empty slack between label and restore `+` doubled as a project-restore click target. Right-aligning occupies most of it.

Verified: 46 chip/minimize DOM tests pass, `biome ci .` clean. Not visually checked in a browser.

### DIFF
--- a/.changesets/minimized-chip-count-alignment.md
+++ b/.changesets/minimized-chip-count-alignment.md
@@ -1,0 +1,9 @@
+---
+issue: null
+pr: null
+type: changed
+bump: patch
+---
+- On a minimized project chip in the sidebar, the session count is now
+  right-aligned instead of sitting next to the project name, with the
+  attention badge immediately to its left. Counts line up down the list.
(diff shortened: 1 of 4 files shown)


lucascaro:
Pushed a test update: the CI failure was `minimize-project.spec.ts`, whose guard asserted the strip left of the restore `+` is dead space — the spec-255 tradeoff called out above, not a Playwright/version problem. Replaced it with the behaviour spec 255 actually promised (clicking there restores, every node lives inside `.hv-chip__open`) plus a new case pinning count-flush-right / attention-to-its-left. Verified the new case fails without the label's `flex: 1`.

PR #361, opened 2026-09-06
fix(daemon): per-user socket dir + events-only socket for agent children

lucascaro:
Closes findings 1 and 2 of the 2026-09-05 security audit.

## Finding 1 — socket squatting

`SocketPath()` returned `/tmp/hive-<uid>/hived.sock` on macOS, and on Linux without `XDG_RUNTIME_DIR`. `EnsureSocketDir` was `MkdirAll(dir, 0700)`, which is a no-op on a directory that already exists — so another local account could pre-create it, park a socket there, and have the GUI hand an impostor every keystroke.

- The macOS default is now `$TMPDIR/hive/` (launchd's private per-user temp dir); Linux falls back to the state dir instead of `/tmp`. `HIVE_SOCKET` still overrides.
- `CheckSocketDir` refuses a socket directory that is a symlink, is not owned by this uid, or carries any group/other permission bit. `EnsureSocketDir` creates 0700 and then verifies.
- The daemon checks before binding; the GUI (`dialOrSpawn`) and hivebar (`Client.session`) check before dialing.

## Finding 2 — `HIVE_SOCKET` is the control socket

Every spawned session inherits `HIVE_SOCKET`. It named the control socket, so any subprocess of an agent could send `HELLO{mode:create, cmd:[...]}` and run arbitrary commands outside the agent's sandbox — or kill sessions, remove worktrees with `force`/`deleteRemote`, or shut the daemon down.

- The daemon binds a second listener at `<sock>.events`. It serves `HELLO{mode:event}` alone and answers every other mode with a new `mode_not_allowed` error frame.
- `registry.SetSocketPath` is now given that path, so `hiveEnv` exports it. `hived hook` and the Pi extension speak `ModeEvent` only and are unchanged.
- The control socket is untouched and still serves all four modes.

Finding 9 (an agent can spoof its own hook events, since it knows `HIVE_SESSION_ID`) stays accepted — this reduces the blast radius to state display.

## Contract

`DaemonContract` 5 → 6. A session revived by an older daemon is handed the control socket, so the property only holds once the daemon is the new one.

## Verification

`go test ./... -count=1`, `go vet ./...`, `scripts/check-daemon-contract-selftest.sh`. Plus a live run of the built `hived`:

- both sockets appear in a 0700 dir;
- `mode:create` on the events socket returns `mode_not_allowed` and creates nothing;
- `mode:control` on the control socket still returns a WELCOME (`daemon_contract: 6`);
- `hived` in a 0777 socket dir exits with `has mode 0777; must be 0700`.

Pre-existing and unrelated: `internal/registry` `TestTerminalQueriesAreNotWork` also fails on a clean `origin/main`.

### DIFF
--- a/.changesets/socket-hardening.md
+++ b/.changesets/socket-hardening.md
@@ -11,14 +11,14 @@ bump: minor
   shared machine another account could previously plant a fake socket
   there and read everything you typed.
 - Programs running inside a Hive session can now only report their state
-  back to Hive and capture ideas for their own project (`hive idea`). They
+  back to Hive and capture ideas for their own project (`hived idea`). They
   could previously use the same connection to create, attach to or kill
   sessions, remove worktrees, shut the daemon down, or list every session
   you have open. This raises the bar rather than sealing a boundary — an
   agent runs as you and Hive does not sandbox it — so `SECURITY.md` now says
   so plainly. Requires a daemon restart, which Hive will prompt for after
   the update.
-- `hive idea list --all` is refused from inside a session, where `hive idea
+- `hived idea list --all` is refused from inside a session, where `hived idea
   list` now shows that session's project. `--all` works from an ordinary
   shell, which it did not before — it needed a session's environment and
   was therefore unreachable in every context the moment the in-session case
--- a/.changesets/socket-hardening.md
+++ b/.changesets/socket-hardening.md
@@ -11,7 +11,16 @@ bump: minor
   shared machine another account could previously plant a fake socket
   there and read everything you typed.
 - Programs running inside a Hive session can now only report their state
-  back to Hive and capture ideas (`hive idea`). They could previously use
-  the same connection to create, attach to or kill sessions, remove
-  worktrees, shut the daemon down, or list every session you have open.
-  Requires a daemon restart, which Hive will prompt for after the update.
+  back to Hive and capture ideas for their own project (`hive idea`). They
+  could previously use the same connection to create, attach to or kill
+  sessions, remove worktrees, shut the daemon down, or list every session
+  you have open. This raises the bar rather than sealing a boundary — an
+  agent runs as you and Hive does not sandbox it — so `SECURITY.md` now says
+  so plainly. Requires a daemon restart, which Hive will prompt for after
+  the update.
+- `hive idea list --all` is refused from inside a session, where it now
+  lists only that session's project. Run it from an ordinary shell for the
+  cross-project view.
+- Hive refuses to start a second daemon against one state directory. The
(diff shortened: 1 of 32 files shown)


lucascaro:
## Review loop: escalated at max iterations

5 of 5 iterations run. Stopping because the loop budget is spent, not because findings remain unaddressed — every finding raised across the five rounds is either fixed on this branch or explicitly accepted below.

**Head `5076820c` — CI fully green** (Linux, Windows, macOS, CodeQL, `daemon-contract`, both generated-file gates). `MERGEABLE`, zero review threads.

### Trajectory

| Iter | BLOCKING | Outcome |
|---|---|---|
| 1 | 1 | Threat model wrong: verbs allowlisted, data not scoped |
| 2 | 0 | `idea list --all` unreachable everywhere; flock/restart race |
| 3 | 0 | One missing test |
| 4 | 0 | Flock rationale claimed something the code can't do |
| 5 | 0 | Legacy probe skipped `CheckSocketDir` |

The last one was the sharpest: the socket-move migration probe dialled `/tmp/hive-<uid>/hived.sock` without the ownership/mode/symlink check every other leg applies, reintroducing this branch's own `/tmp` squat on a new path. Not impersonation — every client still checks before handshaking — but a denial of service from any local uid, wedging the daemon, hivebar and the GUI at once. Fixed in `5076820c` with one guard in the shared resolver.

### Accepted, with reasons

- **Defence in depth, not a boundary.** The control socket sits beside the events one, same uid; a process running as the user can dial it whatever `HIVE_SOCKET` says. POSIX permissions cannot separate two processes under one uid, and Hive does not sandbox agents. `SECURITY.md` says this outright rather than letting the narrowing read as containment.
- `CheckSocketDir` verifies the immediate parent, not the ancestor chain, and there is a TOCTOU window between check and bind/dial. Every default path's ancestors are per-user 0700 (launchd `$TMPDIR`, `$XDG_RUNTIME_DIR`, `$HOME`).
- `HIVE_SOCKET` is reused rather than split into a second variable, so `hived hook` and the Pi extension keep working unchanged.
- `ModeEvent` still trusts the payload's `session_id` — audit finding 9, accepted; this PR's purpose there is to limit its blast radius to state display.
- The legacy-path migration is one release of transitional code. `LegacySocketPath`'s comment says to delete it and its callers once nobody is upgrading across the move.

### Not from this PR

`internal/registry` `TestTerminalQueriesAreNotWork` fails on a clean `origin/main` too (verified by stashing). CI's Go-unit leg does not appear to run it.

### Suggested next step

Human review of the diff, then merge. Merging is deliberately not done from the loop.

PR #377, opened 2026-09-07
feat(ideas): start a session from an idea (337 phase 3)

lucascaro:
Phase 3 of 3 for the idea inbox (spec `docs/product-specs/337-idea-inbox.md`,
plan `docs/exec-plans/active/337-idea-inbox.md`). Phase 1 shipped in #352,
phase 2 in #358.

## What this adds

**Start a session from an idea.** Every open note in the inbox gets a
*Start session* action: the agent launcher opens with the project pinned
and an opening prompt built from the note, worktree checkbox honoured.

**The prompt is an instruction, not a label.** The plan specified
`"<Kind>: <text>"`; raised on first read, an agent handed
`Idea: the grid loses focus` knows what was noticed and nothing about
what to do with it, so it guesses — usually by editing code off a
one-line note that was never a specification. `ideaPrompt` picks a verb
from the kind (bug: reproduce, root-cause, report back before touching
code; idea: explore, ask, propose a plan; feedback: assess and
recommend) and passes the note through verbatim as the subject. One
line per kind, because the typed-delivery path appends a carriage
return and TUIs disagree about whether an embedded newline submits
early.

**The prompt is editable in the launcher.** The note was jotted down
mid-task against a different problem; this is the last moment to turn
it into a brief. Edits do not write back to the idea — the record is
what was noticed, this is the brief for one session.

**Delivery.** Claude and Pi take it as an argv positional
(`agent.Def.PositionalPrompt`, verified interactive under a real PTY on
2026-09-07). Every other agent has it typed into the PTY on the first
idle edge *after* the session has been seen working — not the first
idle, which is t=0, before the TUI has drawn an input box to type into.

**Re-kind and re-project.** `UPDATE_IDEA` gains `kind` and
`project_id`, validated by the registry against `wire.IdeaKinds` and
the live project map, and the inbox's Edit reuses the ⌘I capture sheet
in an edit mode. The capture sheet pre-fills the project from whatever
session was focused, so a mis-filed note is the default behaviour being
wrong rather than user error.

## Two deliberate departures from the plan

**`CreateSpec.IdeaID` is added, reversing the 2026-09-05 decision that
rejected it.** That decision assumed the GUI could send
`UPDATE_IDEA{status:started, session_id}` after the create returned —
but the GUI never learns the new session's id. `CREATE_SESSION` is
fire-and-forget over the control connection, the daemon mints the uuid,
and nothing correlates the `SESSION_EVENT(added)` that follows with the
request that caused it. The rejected alternative was not "a smaller
crash window at the cost of a second create path" — it was the only
path there is. One `omitempty` string plus an in-memory `Entry.ideaID`
is the whole cost, and it makes the plan's "flip the idea only after
the prompt is delivered" constraint implementable rather than
aspirational.

**No daemon-level `TestCreateSessionWithInitialPrompt`.** The
`CREATE_SESSION` arm decodes `wire.CreateSpec` and hands it to
`registry.Create` untouched; there is no daemon-side behaviour left
that the wire round-trip tests and the registry's delivery tests do not
already cover. The plan's `TestRestartDoesNotResendPrompt` is realized
as `TestPositionalPromptIsAppendedOnlyForItsAgents` (the prompt joins
argv in `resolveAgentCmd`, which restart never calls) plus
`TestPendingPromptDeliveredOnlyOnce`.

## Contract

`DaemonContract` 6 → 7. Both new directions are silently *wrong* rather
than silently empty against an older daemon: it drops the unknown
`initial_prompt` / `idea_id` / `kind` / `project_id` JSON fields, so a
newer GUI would show a session that never got its prompt and a re-kind
that never happened.

`App.CreateSession` also converts from 11 positional parameters to a
single `CreateSessionOpts` struct, per the plan — the four call sites
were being touched anyway.

## Two layout bugs the jsdom suite could not see

Both found by measuring in Chromium, both green across all 1174 jsdom
assertions:

- the opening-prompt field overflowed the launcher popup by 14px (354px
  inside a 350px popup) — this theme sets `box-sizing` per rule rather
  than globally, so `width: 100%` plus padding and border overran;
- the popup is shrink-to-fit at `min-width: 220px`, right for a filter
  box and cramped for a three-sentence brief.

The bounding-box comparison is kept as a Playwright test rather than
thrown away.

## Verification

`go build ./...`, `go vet ./...` and `staticcheck ./...` for
darwin/linux/windows, `go test ./...`, the isolated
`go test -tags=e2e ./cmd/hived/...` (temp `HOME`/`HIVE_STATE_DIR`/
`HIVE_SOCKET`), `npm run typecheck`, `biome ci .`, `vitest run` (1174),
the full Playwright mock suite (287 passed / 31 skipped),
`check-changeset.sh`, `check-daemon-contract.sh` (6 → 7 OK),
`ui-lint.sh --strict` and `--contrast`.

Two failures are **pre-existing on `main`** and untouched by this diff,
confirmed by running them in a throwaway worktree at `867d1ab2`:
`TestTerminalQueriesAreNotWork` (deterministic locally) and
`GOOS=windows staticcheck`'s `internal/daemon/lock.go:28 stateLockPoll
is unused`.


lucascaro:
## Review loop stopped: budget reached (5/5 iterations)

Not a convergence — the loop ran out of iterations, so **the current
head has not itself been reviewed**. Recording where it stopped.

### Trend

| Iter | Verdict | BLOCKING | IMPORTANT | Threads |
|---|---|---|---|---|
| 1 | REQUEST_CHANGES | 0 | 5 | 0 |
| 2 | REQUEST_CHANGES | **1** | 7 | 0 |
| 3 | COMMENT | 0 | 2 | 0 |
| 4 | REQUEST_CHANGES | 0 | 8 | 0 |
| 5 | COMMENT | 0 | 5 | 0 |

Every finding above MINOR was fixed rather than left standing; each is
recorded in the exec plan's Decision log with its reasoning. Zero
review threads were open at any point.

### What the loop actually caught

The two that justify the whole exercise, both reproduced before being
fixed rather than argued about:

- **iter 2, BLOCKING** — a note started as a session with the **Shell**
  agent was typed into a bare shell and submitted, i.e. executed. Shell
  is the launcher's default-selected first row, so Start session →
  Enter was the *default* route. A note containing `$(touch <marker>)`
  created the marker. Fixed by making delivery an explicit opt-in
  (`Def.TypedPrompt`) and collapsing the three-way decision into one
  function, `deliveryFor`.
- **iter 5, SECURITY** — on Windows the argv prompt reaches
  `cmd.exe /S /C`, and `cmdExeEscape` documents that it does *not*
  escape `%`. A note reading `%GITHUB_TOKEN%` would have been expanded
  out of the daemon's environment into the agent's first turn.

Two rounds of findings were consequences of the *previous* round's fix
(iter 3 and half of iter 5), which is the argument for having run it to
five rather than stopping at the first green-looking pass.

### Also worth flagging

- Three findings were bugs in **this PR's own tests**, including a
  Playwright e2e that asserted the opposite of the shipped daemon rule
  and stayed green because the mock ignored the capability it was
  testing.
- Two CI failures turned out not to belong to this PR:
  [#379](https://github.com/lucascaro/hive/issues/379) (Linux
  `Session.Done()` never closes when a child exits by itself —
  reproduced against a clean `origin/main`) and spec 374's
  `stage: BACKLOG`, which had been failing `verify-generated` on *every*
  PR in the repo and is fixed here.

### Suggested next step

The head has moved since the last review. Either re-run
`/hs-review-loop 377` for a confirmation pass over
`569aadef`, or go straight to `/hs-merge-gate 337` and accept that the
final round of fixes is unreviewed. The spec stays at `REVIEW` — this
run did not earn `GATE`.

PR #387, opened 2026-09-10
fix(gui): stop Ctrl+Shift+V pasting twice

qmertesdorf:
## The bug

`Ctrl+Shift+V` in a session terminal pastes the clipboard **twice**.

## Root cause

`SessionTerm`'s custom key handler (`cmd/hivegui/frontend/src/app/session-term.js`) owns `Ctrl+Shift+V`: it reads the system clipboard over the Wails bridge, writes the text to the PTY, and returns `false`.

Returning `false` from `attachCustomKeyEventHandler` only tells xterm.js to skip its own *keydown* translation — it does **not** cancel the DOM default action (xterm's keydown listener discards the return value of `_keyDown`, and the early `return false` happens before xterm's own `preventDefault()`).

So the webview went on to perform its native paste. That fires a `paste` event on xterm's helper textarea, where xterm registers its own listener (`handlePasteEvent` → `triggerDataEvent` → `onData` → `_writePty`). The same clipboard text reached the PTY a second time.

One keypress, two writes.

The two sibling branches in the same handler (Cmd+Backspace, Shift+Enter) already call `e.preventDefault()`; the `Ctrl+Shift` block was the odd one out.

## The fix

1. `e.preventDefault()` on the branches that consume the key, so the webview cannot run a native paste on top of ours.
2. The clipboard write now goes through `term.paste()` instead of a raw `_writePty()`.

(2) is part of the same fix, not a drive-by: suppressing the native path also removes the CRLF→CR normalisation and bracketed-paste framing (DECSET 2004) that xterm applied. Those arrived *via the duplicate*. Without this, killing the duplicate would leave only the raw write, and every multi-line paste would submit once per line in Claude/Codex.

## Tests

New `test/dom/paste.test.js` drives the real handler through a real xterm instance in jsdom:

- **`cancels the keydown default…`** — the bug in one assertion (`ev.defaultPrevented`). Fails before the fix.
- **`writes the clipboard exactly once`**
- **`normalises newlines…`** and **`brackets the paste when the running program enabled DECSET 2004`** — both fail before the fix.
- **`xterm native paste path (evidence for the duplicate)`** — dispatches a bare `paste` event with no keydown and shows it writes to the PTY on its own. This is the second write the user saw; it is reachable only while the keydown default is left intact.

Pre-fix: 3 failed / 2 passed. Post-fix: 5 passed.

Full frontend suite: **26 files, 203 tests, all passing.** `vite build` clean.

https://claude.ai/code/session_01SPt4Ra6ndrNVYG1jMhNRTB

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `Ctrl+Shift+V` so clipboard content is pasted only once.
  - Preserved newline normalization and bracketed-paste behavior for multi-line text.
  - Prevented unintended browser actions for terminal copy, paste, and select-all shortcuts.
- **Tests**
  - Added coverage for single-paste behavior, clipboard handling, newline normalization, and bracketed paste.

### DIFF
--- a/.changesets/paste-fires-twice.md
+++ b/.changesets/paste-fires-twice.md
@@ -1,0 +1,12 @@
+---
+type: fixed
+bump: patch
+pr: 387
+---
+
+`Ctrl+Shift+V` pastes once instead of twice. The terminal read the clipboard and
+wrote it to the session itself, but never cancelled the keypress, so the webview
+also ran its own paste on top — the same text arrived twice on every use. The
+paste now also goes through xterm's paste path, so multi-line clipboard content
+keeps its newline normalisation and bracketed-paste framing instead of being
+written raw, which had agents submitting once per pasted line.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,13 @@ navigation, ⌘1-9, the tray and the command palette now all follow the order th
 rows are actually painted in.
 
 ### Fixed
+`Ctrl+Shift+V` pastes once instead of twice. The terminal read the clipboard and
+wrote it to the session itself, but never cancelled the keypress, so the webview
+also ran its own paste on top — the same text arrived twice on every use. The
+paste now also goes through xterm's paste path, so multi-line clipboard content
+keeps its newline normalisation and bracketed-paste framing instead of being
+written raw, which had agents submitting once per pasted line.
+
 Confirmation dialogs now work on Windows. Deleting a project, killing a live
 session, restarting Hive and applying an update all go through a native
 confirmation, and every one of them silently did nothing on Windows: the dialog
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,13 +20,6 @@ navigation, ⌘1-9, the tray and the command palette now all follow the order th
 rows are actually painted in.
 
 ### Fixed
-`Ctrl+Shift+V` pastes once instead of twice. The terminal read the clipboard and
-wrote it to the session itself, but never cancelled the keypress, so the webview
-also ran its own paste on top — the same text arrived twice on every use. The
-paste now also goes through xterm's paste path, so multi-line clipboard content
-keeps its newline normalisation and bracketed-paste framing instead of being
-written raw, which had agents submitting once per pasted line.
-
 Confirmation dialogs now work on Windows. Deleting a project, killing a live
 session, restarting Hive and applying an update all go through a native
 confirmation, and every one of them silently did nothing on Windows: the dialog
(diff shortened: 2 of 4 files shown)


lucascaro:
## Review

**Scope:** 3 files · 228 changed lines · base `main`.
**Investigated beyond the diff:** convention drift (changeset/CHANGELOG pipeline), test reach (ran the unit+dom suite on this PR head, macOS), and the macOS paste path (`cmd/hivegui/menu_darwin.go` Edit menu → WKWebView `paste:`).
**Skipped:** call sites (no exported signature changed), implementations (no interface touched), deletions (nothing removed), consumers (no schema change), supply chain (no dependency change).

The root-cause analysis is right and the fix is the correct one: `preventDefault()` on the branches that consume the key, and routing our own write through `term.paste()` so the surviving write keeps the CRLF folding and DECSET 2004 framing the suppressed native path used to provide. One blocking issue, and it is in the changelog, not the code.

### Does it work on macOS?

Yes, and there is no regression there.

- `test/dom/paste.test.ts` passes 7/7 on macOS.
- Mac users paste with ⌘V, which never reaches this handler — the guard `if (!e.ctrlKey || !e.shiftKey || e.altKey || e.metaKey) return true` bails on `metaKey`. ⌘V comes from `menu.EditMenu()` (`cmd/hivegui/menu_darwin.go:100`) → native `paste:` on the webview → DOM `paste` event → xterm's own paste listener → `onData` → `_writePty`. This diff does not touch that path, so it stays a single write.
- The double-paste was therefore Ctrl+Shift+V-only, and macOS already had newline normalisation and bracketed-paste framing via that native path. This PR makes the Ctrl+Shift+V path match what ⌘V already did, which also removes an asymmetry between the two.
- Ctrl+Shift+V is not `isMac`-gated, so mac gets the fix as well.

Full unit+dom run on the PR head shows 14 dom failures, all `Failed to resolve import "../wailsjs/go/main/App"` — missing generated Wails bindings in a fresh worktree, unrelated to this diff.

### BLOCKING

**`CHANGELOG.md:22` — hand-edits a generated region, and the branch has no `.changesets/*.md` entry.**

`scripts/regen-generated.py` regenerates the `## [Unreleased]` body *from* `.changesets/*.md`. This PR writes prose directly into that generated block and adds no changeset, so two things happen:

1. `scripts/check-changeset.sh` and the `changesets` workflow fail the branch — merge is gated.
2. The next `chore: regenerate aggregates` commit overwrites the block, so the change ships with no changelog line at all.

Fix: drop the `CHANGELOG.md` hunk and add `.changesets/paste-fires-twice.md` with the same prose:

```markdown
---
type: fixed
bump: patch
---

`Ctrl+Shift+V` pastes once instead of twice. …
```

Then let `regen-generated.py` produce the CHANGELOG body.

### Minor

`test/dom/paste.test.ts` — the `defaultPrevented` assertion is a proxy for the real defect: jsdom never performs a native paste, so no test observes two writes, and the "evidence for the duplicate" test will keep passing regardless of the fix. The PR body says this plainly; noting it only so nobody later reads it as end-to-end coverage. It is still the right assertion to have — it fails the moment someone deletes the `preventDefault()`.

### Verdict

Request changes — code is correct and mac-safe; the changelog just needs to move from the generated file into `.changesets/`.

PR #388, opened 2026-09-11
fix: stop sessions inheriting NO_COLOR from the daemon

qmertesdorf:
## Problem

Every session Hive spawns gets the daemon's environment verbatim, with only `TERM` overridden:

```go
cmd.Env = append(os.Environ(), "TERM=xterm-256color")
```

That makes `NO_COLOR` contagious. If whatever shell launched `hivegui` had it set, `hived` inherits it, and from then on every agent and every plain shell in every tile renders monochrome — for the life of the daemon, with nothing in Hive's own settings to explain it.

I hit this on Windows: a Claude session with no colour at all, while `TERM` was correctly `xterm-256color`. `NO_COLOR` was in neither the User nor Machine registry environment, no shell profile set it, and the string does not appear in Hive's source or in the `hived` binary. Reading the process environment blocks up the tree showed `NO_COLOR=1` present from `hivegui` down through `hived` to the session — inherited once at GUI launch and propagated from there.

## Fix

Extract the environment construction into `sessionEnv()` and drop `NO_COLOR` there. The justification is the one that already applies to `TERM`: a tile is a colour-capable xterm.js terminal whatever the GUI itself was started under, so a value the daemon picked up by inheritance says nothing about it. An explicit `Options.Env` still wins, since it is appended afterwards.

## Testing

Unit tests in `internal/session/env_test.go` cover `TERM` being forced, `NO_COLOR` being dropped, and unrelated variables surviving.

End-to-end, spawning a real session via `session.Start` from a parent holding `NO_COLOR=1` and echoing both variables inside the session:

```
before (main):  MARK_NO_COLOR=[1]  MARK_TERM=[xterm-256color]
after  (this):  MARK_NO_COLOR=[]   MARK_TERM=[xterm-256color]
```

`scripts/test.sh go` is green apart from `TestManagedPath`, which fails identically on unmodified `main` on this machine — it needs the Windows symlink privilege.

https://claude.ai/code/session_014rBWb7d5MqmBcUvySKwMCh

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed monochrome session rendering when the application is launched from a shell with `NO_COLOR` enabled.
  - Sessions now consistently use color-capable terminal settings while preserving other environment variables.

- **Tests**
  - Added coverage for terminal color handling, forced terminal settings, and preservation of unrelated environment variables.

### DIFF
--- a/.changesets/session-strips-no-color.md
+++ b/.changesets/session-strips-no-color.md
@@ -1,0 +1,13 @@
+---
+type: fixed
+bump: patch
+---
+
+Sessions no longer render monochrome when the GUI happened to be launched from
+a shell with `NO_COLOR` set. Hive built every session's environment from the
+daemon's own, overriding only `TERM`, so a `NO_COLOR` that reached `hivegui` by
+inheritance was handed down to every agent and shell in every tile and stayed
+there for the life of the daemon — with no setting anywhere to explain it. A
+tile is a colour-capable xterm.js terminal whatever the GUI was started under,
+so the variable is now dropped when a session's environment is built, for the
+same reason `TERM` is already forced.
--- a/.changesets/session-strips-no-color.md
+++ b/.changesets/session-strips-no-color.md
@@ -1,6 +1,7 @@
 ---
 type: fixed
 bump: patch
+pr: 388
 ---
 
 Sessions no longer render monochrome when the GUI happened to be launched from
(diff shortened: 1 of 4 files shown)


lucascaro:
BTW added the `daemon-contract-override` label so selfupdate doesn't forcea full restart, but you can via the menu after updatign in file->restart daemon

PR #389, opened 2026-09-11
chore: THROWAWAY probe for actions/cache v6 + upload-artifact v7 (do not merge)

lucascaro:
**Do not merge. Do not review.** This branch exists only to exercise two upgrade paths that #372 and #370 cannot reach in their own CI runs, then it gets deleted.

- `actions/cache` v6.1.0 with both cache keys prefixed `v6probe-`, forcing a primary-key miss so the **save** path runs. #372's CI hit every key, which proved restore-from-v4-written-cache only.
- `actions/upload-artifact` v7.0.1 plus a Linux-only step that writes files into `test-results/` and exits 1, so the `if: failure()` upload step actually uploads a real artifact.

The Linux leg is expected to be red. That is the point.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -362,16 +362,6 @@ jobs:
           CI: 'true'
           TERM: dumb
 
-      - name: PROBE (throwaway) force artifact upload path
-        if: matrix.label == 'Linux'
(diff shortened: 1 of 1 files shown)


lucascaro:
Probe done, evidence collected — see the dependabot PRs. Deleting the branch.

PR #394, opened 2026-09-11
fix: stop Windows opening a console window per child process

qmertesdorf:
## What

On Windows, ordinary use of Hive scattered console windows across the screen — a stray popup on every poll, a volley when checking for an update.

`hivegui.exe` is a GUI binary and `hived.exe` is started detached (`DETACHED_PROCESS`), so **neither process owns a console**. When a process with no console starts a console-subsystem child, Windows hands that child a brand new console *window*. Hive shells out constantly: `git` behind worktree inventory and the update check, `gh pr list` for merged branches, `claude --version`, `tasklist`, the staged bundle's `hived --version`. Every one of them was a popup.

`internal/proc` is now the single constructor for non-PTY children. `proc.Command` / `proc.CommandContext` mark the child `CREATE_NO_WINDOW` (plus `SW_HIDE`, for a child that calls `AllocConsole` itself) on Windows, and are a pass-through everywhere else. All 16 spawn sites route through it — including the `tasklist` probe in `restart_windows.go`, which had been carrying the only hand-rolled `HideWindow` workaround.

PTY children are untouched: `internal/session` attaches those to a ConPTY, which never has a window.

## Keeping it fixed

`TestNoDirectExecOnWindows` (`internal/proc/edict_test.go`) parses every file that builds for `windows/amd64` and fails on a direct `exec.Command` / `exec.CommandContext`. A direct call compiles and runs perfectly well while putting the popups back for every Windows user, so nothing but a test stops the next spawn site from regressing this.

Three exemptions, each with its reason in the allowlist:

| File | Why |
|------|-----|
| `cmd/hivegui/window_windows.go` | `DETACHED_PROCESS` already leaves the child consoleless, and `CREATE_NO_WINDOW` may not be combined with it — `CreateProcess` rejects the pair |
| `cmd/hivegui/spawn_windows.go` | same, routes through `startDetachedWindows` |
| `cmd/hivegui/os_terminal.go` | opening a visible terminal is the point |

Written up as a hard rule in `DESIGN.md` and a change pattern in `AGENTS.md`.

## Test plan

- `TestNoDirectExecOnWindows` went red listing all 16 violations, then green.
- New unit tests assert the flags are set by both constructors, that `HideConsole` adds to a caller's existing `CreationFlags` rather than replacing them, and that the wrappers build the same `Path`/`Args` as `os/exec`.
- Full `go test ./...` on windows/amd64: no new failures. Three pre-existing environment failures are unchanged (`TestManagedPath` needs symlink privilege, `TestPackageIsIsolatedFromRealHiveState` trips because `%TEMP%` lives under `%LOCALAPPDATA%`, `TestEmbeddedAssetsIncludeWebfonts` needs a frontend build) — identical with the change stashed.
- Cross-compiles clean for linux/amd64 and darwin/arm64; `go vet` clean; `./build.sh --platform windows` builds both binaries.
- **Measured on Windows 11.** A GUI-subsystem probe (no console, same situation as `hivegui`/detached `hived`) spawning `cmd /c ping`, diffing visible top-level windows via `EnumWindows` + `IsWindowVisible`:

  ```
  plain exec.Command : 2 new visible windows (WindowsTerminal, cmd)
  with the flags     : 0 new visible windows
  ```

  Worth noting for anyone verifying this by hand: `MainWindowHandle` and the conhost process count do **not** discriminate — `CREATE_NO_WINDOW` still allocates a console, it just never shows one. Only enumerating visible windows does.

https://claude.ai/code/session_014rBWb7d5MqmBcUvySKwMCh

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Windows console windows flashing during background operations and helper commands.
  - Git, GitHub CLI, Claude version checks, update checks, and related background tasks now run without visible console popups.
  - Existing command behavior, output, timeouts, and error handling remain unchanged.

- **Documentation**
  - Added guidance documenting background process behavior and command execution requirements.

- **Tests**
  - Added coverage to verify hidden Windows processes and prevent regressions.


lucascaro (commented):
## Review: mac-safety + enforcement test

**Does not break macOS.** Verified by building and running, not by reading:

- `hide_other.go` is `//go:build !windows` with an empty `HideConsole`, so `SysProcAttr` stays `nil` on darwin — every converted site builds the identical `exec.Cmd` it did before.
- `go vet ./cmd/... ./internal/...` clean on darwin **and** `GOOS=windows`; `go build ./...` clean on both.
- `go test` green for `internal/proc`, `internal/worktree`, `internal/registry`, `internal/agent`, `cmd/hivegui` (the only failures were `TestEmbeddedAssets*`, from my stubbed `frontend/dist`, not this branch).
- The `os/exec` import kept in `update_latest.go` is still used (`*exec.ExitError`, line 35) — not a dangling import.

Darwin-only spawn sites (`update_apply_darwin.go`, `update_verify_darwin.go`, `shell_env_darwin.go`, `window_unix.go`, `cmd/hivebar`, `menubar_darwin.go`) are correctly left alone — they don't build for Windows.

### IMPORTANT

**`internal/proc/edict_test.go:60` — `TestNoDirectExecOnWindows` walks into nested checkouts and fails on other branches' code.**

The walk skips `.git`, `node_modules` and `.worktrees`, but not `.claude/worktrees/`, which Claude Code creates inside the repo root — nor any other nested clone. Reproduced against the maintainer's main checkout: the test reports 19 bogus violations sourced from `.claude/worktrees/wails-version-check/...`, i.e. the pre-fix code on an unrelated branch. CI is fresh-cloned so it stays green, but `go test ./...` locally goes red permanently, and the red is unreadable noise — which is exactly how an edict test gets ignored.

Fix (verified: reruns clean with it):

```go
switch d.Name() {
case ".git", "node_modules", ".worktrees":
	return fs.SkipDir
}
if path != root {
	if _, err := os.Stat(filepath.Join(path, ".git")); err == nil {
		return fs.SkipDir // nested checkout / worktree
	}
}
```

Skipping any directory carrying a `.git` covers `.claude/worktrees`, `.worktrees` and stray clones in one rule — better than extending the hardcoded name list.

### MINOR

- `internal/proc/edict_test.go:109` — `import . "os/exec"` bypasses the scanner: `local` becomes `"."`, the calls parse as `*ast.Ident`, and nothing matches. (Same gap CodeRabbit flagged, still open.) Negligible in practice, two lines to close: treat `local == "."` like `"_"`, or reject the dot import outright.
- `internal/proc/proc_test.go:41` — `TestHideConsoleIsIdempotent`'s comment claims it covers "a `*exec.Cmd` built elsewhere, including one that already carries `SysProcAttr`", but it only calls `proc.Command`, and off Windows it asserts nothing at all. The real coverage is `TestHideConsolePreservesExistingFlags` (windows-only). Fix the comment, or construct the `exec.Cmd` directly.

### Verdict

**COMMENT** — the fix is correct and genuinely inert on macOS. The enforcement test needs the nested-repo guard, or it cries wolf on the primary checkout from day one.

internal/proc/edict_test.go
```diff
@@ -0,0 +1,166 @@
...
+		if !d.IsDir() {
+			return nil
+		}
+		switch d.Name() {
+		case ".git", "node_modules", ".worktrees":
+			return fs.SkipDir
+		}
+		pkg, err := bctx.ImportDir(path, 0)
+		if err != nil {
+			// No buildable Go in this directory for windows/amd64.
```
lucascaro, line 63: **IMPORTANT — this walk descends into nested checkouts and fails on other branches' code.**

`.git`, `node_modules` and `.worktrees` are skipped, but `.claude/worktrees/` (created by Claude Code inside the repo root) is not, and neither is any other nested clone. Reproduced against the maintainer's main checkout: 19 bogus violations, all sourced from `.claude/worktrees/wails-version-check/...` — the pre-fix code on an unrelated branch. CI is fresh-cloned so it stays green, but local `go test ./...` goes permanently red with unreadable noise, which is how an edict test ends up ignored.

Verified fix (reruns clean with it):

```go
switch d.Name() {
case ".git", "node_modules", ".worktrees":
	return fs.SkipDir
}
if path != root {
	if _, err := os.Stat(filepath.Join(path, ".git")); err == nil {
		return fs.SkipDir // nested checkout / worktree
	}
}
```

Skipping any directory carrying a `.git` covers `.claude/worktrees`, `.worktrees` and stray clones in one rule — better than extending the hardcoded name list.

internal/proc/edict_test.go
```diff
@@ -0,0 +1,166 @@
...
+	local := ""
+	for _, imp := range f.Imports {
+		p, err := strconv.Unquote(imp.Path.Value)
+		if err != nil || p != "os/exec" {
+			continue
+		}
+		local = "exec"
+		if imp.Name != nil {
+			local = imp.Name.Name
+		}
```
lucascaro, line 111: **MINOR — `import . "os/exec"` bypasses the scanner.** With a dot import `local` becomes `"."`, the constructor calls parse as `*ast.Ident` rather than `*ast.SelectorExpr`, and nothing matches. (Same gap CodeRabbit flagged, still open.) Negligible in practice, two lines to close: treat `local == "."` like `"_"`, or reject the dot import outright.

internal/proc/proc_test.go
```diff
@@ -0,0 +1,45 @@
...
+	cmd := proc.CommandContext(ctx, "git", "status")
+	if err := cmd.Start(); err == nil {
+		_ = cmd.Wait()
+		t.Error("Start() on a cancelled context succeeded; context not wired through")
+	}
+}
+
+// HideConsole must be safe on a *exec.Cmd built elsewhere, including
+// one that already carries SysProcAttr.
+func TestHideConsoleIsIdempotent(t *testing.T) {
```
lucascaro, line 41: **MINOR — the comment overstates what this test covers.** It claims "a `*exec.Cmd` built elsewhere, including one that already carries `SysProcAttr`", but the test only calls `proc.Command`, and off Windows `HideConsole` is a no-op so it asserts nothing at all. The real coverage is `TestHideConsolePreservesExistingFlags` (windows-only). Fix the comment, or construct the `exec.Cmd` directly here.

lucascaro (commented):
Inline notes for the findings in my review above.

lucascaro:
Moved to #400 (same commits, now on a branch in this repo). Thanks @qmertesdorf — authorship is preserved.

internal/proc/edict_test.go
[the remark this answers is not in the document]
lucascaro: Fixed in 9283665a: dot-imported os/exec is matched as a bare identifier; caf7fc76 also tracks every local name so duplicate imports can't hide a call. Work continues in #400.

internal/proc/edict_test.go
lucascaro: Fixed in 9283665a: scanTree skips any directory carrying a .git entry (clone or worktree), with a fixture test; caf7fc76 adds a scanned-file floor so an over-broad skip can't pass vacuously. Work continues in #400.

internal/proc/edict_test.go
lucascaro: Fixed in 9283665a: dot-imported os/exec is matched as a bare identifier; caf7fc76 also tracks every local name so duplicate imports can't hide a call. Work continues in #400.

internal/proc/proc_test.go
lucascaro: Fixed in 9283665a: test now builds the cmd with exec.Command, compares SysProcAttr values, and asserts nil off Windows. Work continues in #400.

PR #399, opened 2026-09-13
feat(gui): in-app update on Windows

qmertesdorf:
> **Stacked on #394.** This branch contains that PR's two commits, so the diff
> below includes them. Land #394 first; this is the rest.

## What this does

In-app update now works on Windows, on both channels. Windows previously got
"download it manually on this platform" — advice that is simply wrong on the
latest channel, which tracks a git checkout and has no release artifact to
download at all.

Replacing a running `hivegui.exe` needs no helper process and no elevation.
Windows refuses to overwrite or delete a mapped image but allows renaming one,
so the update renames the running binaries aside, installs over them, and sweeps
the displaced copies at the next start. A failure part-way through rolls back and
leaves a launchable Hive.

Design notes and the staging/rollback sequence are in
`docs/exec-plans/active/windows-in-app-update.md`.

## Never hand build.sh a WSL bash

`findBash` took the first `bash` on the PATH. On any machine with WSL installed
that is a distro launcher — `System32\bash.exe`, or the Store app execution alias
in `WindowsApps` — both of which sit on a default PATH ahead of Git for Windows'
bash.

`buildCommandLine` writes a Windows path deliberately, because Git Bash accepts
one without needing `cygpath`. Inside a distro that path lives under `/mnt`, so a
real update died with:

```
build.sh failed: /bin/bash: line 1: cd: D:/git/hive: No such file or directory
```

and a Linux bash could not have produced a Windows binary even if the path had
resolved. Those two launchers are now skipped, falling through to the Git for
Windows candidates; when no Git Bash is installed the refusal names WSL as the
reason rather than claiming bash is not on the PATH — one *is* on the PATH, it is
just the wrong one, and the old wording sent the user hunting for something they
already had.

## daemon-contract

`internal/registry/ghmerged.go` is in the gate's scope, but the change is
`exec.CommandContext` → `proc.CommandContext` — the console-window-hiding
wrapper, and nothing a client can observe. Applying `daemon-contract-override`
rather than bumping the contract, which would cost every user their running
sessions for a refactor they cannot see.

## Testing

Unit tests: `update_apply_windows_test.go` covers staging, rename-aside install,
rollback and the sweep; `shell_env_windows_test.go` covers the bash selection,
including both WSL launcher locations.

On a real Windows machine: rebuilt and installed to `%LOCALAPPDATA%\Programs\Hive`,
and the update check runs clean end to end — settings load, source repo resolves,
git comparison reports up-to-date.

What has **not** been exercised live is `build.sh` itself on the latest channel
since the WSL fix, because the checkout was at its upstream tip and there was
nothing to build. The bash selection is covered by the tests above, and the
command line was verified by hand against Git Bash, but a full latest-channel
build has not run through this path end to end.

---

## Update after review

Rebased onto the current #394 branch, so this no longer carries stale copies of
the files fixed there.

**An interrupted update was unrecoverable.** Between renaming the old image aside
and moving the new one in, the install directory holds no `<exe>` at all — and
the startup sweep deleted both halves. `hived.exe` was then gone for good, with
no route back from inside the app (the release channel refuses when you are
already on the newest version), and the same window for `hivegui.exe` left Hive
unlaunchable. The sweep now repairs before sweeping, preferring the incoming
image.

**The sweep raced a concurrent swap.** Each GUI window is its own process, so
opening a second one mid-swap deleted `.<exe>.new` under the first and failed it
with a "cannot find the file" error. `.new` is no longer swept; the next swap
overwrites it before use.

**`findBash` gave up on the PATH too early.** It took `LookPath`'s first hit and
stopped if that was a WSL launcher — which it usually is, since `WindowsApps`
sits near the front. Anyone whose Git for Windows is not at a default location (a
scoop shim, a portable Git, another drive) was refused and told to install the
Git they already had: the exact confusing refusal the WSL skip exists to prevent,
just relocated. It now walks the whole PATH and refuses only when every `bash` on
it is a launcher.

Smaller fixes from the same review: the build-tool probe believes only tokens
that are real build tools (`bash -lc` sources the login profile, so an nvm banner
became `build.sh needs Welcome`); `build.sh` failures keep their exit status
rather than reporting a trailing output line that is often unrelated; and
`isDownloadedStaging` moved to the darwin file where its only caller lives — it
was dead code on the Windows leg, the same unused-symbol problem the earlier
split was made to fix.

`staticcheck` is clean for `GOOS=linux` and `GOOS=windows` across `./cmd/...`.

## Summary by CodeRabbit

- **New Features**
  - Added in-app updates for Windows on release and latest channels.
  - Windows updates verify downloaded files with checksums and support rollback and recovery after interruptions.
  - Update availability now explains when an update cannot be applied.

- **Bug Fixes**
  - Improved Windows environment detection and clearer guidance when required tools are unavailable.
  - Latest-channel users are no longer directed to unavailable release pages.

- **Documentation**
  - Updated README guidance to reflect Windows support and Linux’s release-page update flow.


lucascaro:
## Review loop: escalated after iteration 1 of 5

**Why it stopped:** one finding needs a design decision, so the loop didn't apply a fix automatically.

**Fixed in `ef9cbf16`:** the exec-plan frontmatter is now current (PR number, stage, and the dependency on #400).

**Needs a decision** (IMPORTANT, correctness): `cmd/hivegui/update_apply_windows.go:600`, `repairInterruptedSwap`

While `swapExes` runs, each target file is briefly missing: it has been renamed to `.old`, and `.new` hasn't been moved into place yet. If a second Hive window starts in that gap, its startup repair treats the missing file as an interrupted swap and moves `.new` into place itself. The first window's rename then fails with "cannot find the file" and rolls back, leaving the old version running. The gap is only milliseconds, but a failed update is confusing to users.

Options:
1. **Lock file:** `swapExes` holds a lock file in the install directory, and startup repair skips while it exists. This is the most robust option, but a stale lock needs cleanup.
2. **Age threshold:** repair only when `.new` is older than some cutoff, such as 30 seconds. This is simpler, but it's a heuristic.

**Not acted on** (MINOR): on Windows the update is verified by SHA-256 checksum only, with no publisher signature check. The changeset, README and exec plan already say so.

CI is green on `ef9cbf16`, and there are no open review threads.

**Suggested next step:** pick option 1 or 2, apply it, and re-run the review loop.

PR #405, opened 2026-09-13
fix(gui): give Wails the Windows options it never had

qmertesdorf:
## The problem

`cmd/hivegui/main.go` never constructed `options.Windows` at all. Wails guards its
whole Windows block with `if opts := f.frontendOptions.Windows; opts != nil`, so
with it nil, `PutIsZoomControlEnabled` and `PutIsPinchZoomEnabled` were never
called and WebView2 kept its own defaults.

The visible consequence: Ctrl+wheel and pinch apply a **WebView2 page zoom on top
of** the app's own font zoom, desynchronising every xterm terminal.
(`PutAreBrowserAcceleratorKeysEnabled(false)` already ran unconditionally, which
is why Ctrl+`+`/`-` were unaffected.)

## What ships

```go
DragAndDrop: &options.DragAndDrop{DisableWebViewDrop: runtime.GOOS == "windows"},
Windows: &windows.Options{
    IsZoomControlEnabled: false,
    DisablePinchZoom:     true,
    ResizeDebounceMS:     resizeDebounceMS, // 24
    Theme:                windows.SystemDefault, // unchanged: Wails' nil fallback, pinned
},
```

- `options.App` is extracted into `appOptions(...)` so the fields are assertable.
- `DisableWebViewDrop` is Windows-only: there it maps to WebView2
  `AllowExternalDrag(false)`, which spares in-page HTML5 drags (sidebar reorder).
  It is inert on macOS, and on Linux Wails calls `gtk_drag_dest_unset`, whose
  effect on in-page drags is unverified — so Linux is left as it was.
- `EnableDefaultContextMenu` stays false everywhere and `OnResume` stays unset;
  tests pin both (see review thread for why).

Removed during review (f867293): `OnResume` → `control:disconnect`,
`EnableDefaultContextMenu` on Windows, and the unwired `SetWindowTheme` binding.

## Verified

- Option semantics checked against Wails v2.15.0 `internal/frontend/desktop/windows/frontend.go`.
- `window_options_test.go` asserts the options; CI green on Linux, macOS, Windows.

## Not verified

Not run in a real Windows build. `ResizeDebounceMS` is a trailing debounce on
`chromium.Resize()`, so the feel of a continuous resize drag is untested.

## daemon-contract

`cmd/hivegui` only — outside the gate's watched trees.


lucascaro:
Thanks for this. The zoom/pinch fix and the resize debounce look good. Three parts need a reproducer or a change before this can merge:

**1. `OnResume` → `control:disconnect`.** The claim that emitting this event avoids tearing down a healthy connection doesn't hold. The frontend handler (`app/events.ts`) shows the red `reconnecting…` status and calls `reconnectControl()`. That calls `ConnectControl()`, which unconditionally detaches and closes the current control connection before redialing (`app_control.go`). So every wake would drop a working connection and flash an error. Do you have a reproducer showing that sleep/wake actually breaks the local control connection on Windows? If a wake really does kill it, the read loop already emits `control:disconnect` on its own. Unless there's a repro, I'd suggest dropping `OnResume`.

**2. `EnableDefaultContextMenu` on Windows.** It's unclear what this buys. The frontend has no `contextmenu` handler, and xterm paints to a canvas, so right-clicking a terminal will most likely show WebView2's page menu (Back / Refresh / Save as / Print) rather than Copy/Paste. Refresh would reload the page and drop GUI state. Please either explain and verify on a real Windows build what the menu shows over terminals and inputs, or remove it. An app-owned Copy/Paste menu could be a separate PR.

**3. `SetWindowTheme`.** It's bound but nothing calls it, and it only does anything on Windows, where it sets the native titlebar dark or light (Wails' darwin and linux implementations are empty). Please either wire it to the frontend theme change so the titlebar follows the chosen preset, or remove it along with its test.

Also, the changeset states the context-menu Copy/Paste and wake-reconnect behaviour as fact, even though the PR body says neither was verified. It also says file drop is ignored "on every platform", but the setting is inert on macOS. Please align the changeset with whatever ships.

qmertesdorf:
Addressed in f867293 — all four points held up:

- **`OnResume` → `control:disconnect`**: removed, along with `onSystemResume` and its tests. `ConnectControl` detaches and closes unconditionally, so every wake would have dropped a working connection and flashed the reconnecting banner.
- **`EnableDefaultContextMenu`**: removed. Wails already gates the WebView2 menu on `f.debug || EnableDefaultContextMenu`, so release builds have it off today; turning it on would have exposed Back / Refresh / Save as / Print.
- **`SetWindowTheme`**: removed with its test. Wiring it to the theme picker is a frontend change (bridge + Playwright stubs + settings call sites) — a real Windows gap (a chosen Hive Light under a dark OS keeps a dark titlebar) but a separate PR.
- **Changeset**: rewritten to list only what ships — zoom/pinch, resize debounce, file-drop guard — all Windows-only, with `pr: 405`.

Two more while I was in there:

- `Theme: windows.SystemDefault` was not a fix — Wails already falls back to `SystemDefault` when `Windows == nil`. The comment now says so and the changeset no longer claims the titlebar changed.
- `DisableWebViewDrop` is not Linux-inert: Wails runs `gtk_drag_dest_unset(webview)` unconditionally on that flag, and whether WebKitGTK's in-page HTML5 drag (the sidebar reorder) survives losing the GTK drop target is unverified. Gated to `runtime.GOOS == "windows"`, where WebView2's `AllowExternalDrag(false)` is documented to spare in-page drags; macOS confirmed inert.

Tests updated accordingly (`window_options_test.go`); go build/vet/test green, Linux cross-build passes.

Issue #202, opened 2026-05-15
GUI: minimize sessions to remove from grid view; restore from a tray


lucascaro:
## Description
Users currently can't temporarily hide a session from the grid view without killing it. When many sessions are running, the grid (project view and all-sessions view) gets noisy and important sessions get crowded out.

Add the ability to **minimize** a session: it stays alive (process keeps running, output keeps buffering) but is removed from the grid layout. Minimized sessions live in a lightweight UI affordance (e.g. a tray / chip row) from which the user can **restore** them back to the grid.

### Desired behavior
- A natural per-tile control to minimize a session from grid view (project grid and all-sessions grid).
- A visible "minimized" affordance showing count + each minimized session (name, project, status indicator) with a click-to-restore action.
- Switching to a minimized session in single-session mode should still work (minimize is a grid-visibility concern, not a "hidden from the app" concern).
- State persists across grid ↔ single transitions within an app session. Persistence across app restarts is a follow-up question.
- Keyboard affordance is nice-to-have but not required for v1.

### Out of scope
- Killing / archiving sessions (already covered).
- Persisting minimized state across app launches (decide during triage).

PR #408, opened 2026-09-14
fix(gui): hide minimized sessions from the sidebar

### LINKED ISSUE #407: Hide minimized sessions from the sidebar
## Description

A minimized session (e.g. calm-thorn) still renders as a row in the sidebar, but ⌘↑/⌘↓ skips it because keyboard navigation excludes hidden sessions. The result is a visible row that cannot be reached from the keyboard, which reads as a focus bug.

The sidebar should list exactly the sessions ⌘↑/⌘↓ can focus: minimized sessions leave the sidebar and remain reachable through the minimized-session tray (and the command palette).

lucascaro:
## Summary

Fixes #407. A minimized session kept a dimmed row in the sidebar, but ⌘↑/⌘↓ skipped it (#252), so a visible row could not be reached from the keyboard.

- `Sidebar.tsx` `ProjectItem` paints only sessions ⌘↑/⌘↓ can land on. The active session is the exception: its row stays while it is minimized, marked, with the restore control.
- The card header count and attention still include minimized sessions. So does the per-row shared-worktree count, because a hidden session still holds the worktree. Worktree panels wrap painted rows only.
- The tray above the status bar and ⌘K are the restore paths. ⌘1–9 stays positional (#252 non-goal).
- Specs #202, #250 and #252 are amended, and `components.md` is updated.

## Tests

- `test/dom/minimize-project.test.tsx`: 7 new or reworked tests. 6 failed on the old code. They cover the inactive row being removed, the active row being kept, the row leaving when focus moves, restore keeping the original slot, card count and attention, the hidden worktree sharer, and a `navSession` cycle visiting exactly the painted rows.
- `test/e2e/minimize.spec.ts`: minimize from the row → the row is gone and the tray chip shows → ⌘↓ skips the session → the tray restores it.
- `npm test`, `npm run typecheck`, `biome ci .` and the full `CI=1` Playwright suite pass (326 passed, 31 skipped).

## Pixel baselines

This regenerates `sidebar-*` (×6) and `chrome-classic`, the only baselines that render the seeded sidebar. Those files already predated the #385 sidebar redesign and failed on `main` before this change. `settings-classic`, `launcher-*` and `worktrees-*` also fail on `main` under `HIVE_SNAPSHOT=1`, for reasons unrelated to this change. They are left untouched.

## Summary by CodeRabbit

- **Bug Fixes**
  - Minimized sessions no longer appear as dimmed rows in the sidebar.
  - The active minimized session remains visible until switching away.
  - Keyboard navigation and reordering now skip hidden minimized sessions.
  - Minimized sessions remain included in project and shared-worktree counts.

- **Accessibility**
  - Minimized sessions can be restored from the session tray or command palette.

- **Documentation**
  - Updated product and UI documentation to reflect minimized-session behavior.


lucascaro:
## Review loop: escalated (iteration 1/5)

**Why it stopped:** autofix found a fix it classifies as RISKY, because it changes a keybinding's behaviour. That needs a human decision.

**Unresolved finding (IMPORTANT):**
- `cmd/hivegui/frontend/src/lib/worktree-groups.ts:334` [CORRECTNESS]. `clusterReorderOps` picks the ⇧⌘↑/⇧⌘↓ target over every session in the project, minimized ones included. This PR stops painting minimized inactive sessions. When the neighbour is one of those, the keypress swaps order with an invisible row and nothing moves on screen. No test covers it.

**Tension:** spec #252 requires that "⇧⌘↑ / ⇧⌘↓ still reorder across a minimized sibling exactly as before". That criterion was written when the sibling was still visible.

**Suggested next step:** decide between these:
1. Skip non-painted sessions when choosing the reorder slot, keep the daemon moves computed over the full list, add a test, and amend #252.
2. Accept the dead keypress as a known limitation and record it in the #407 decision log.

No review threads are open, and CI was not re-run because nothing changed.

PR #412, opened 2026-09-14
fix(theme): keep a session-only System pair when a store refuses the write

qmertesdorf:
Follow-up to #410, which merged while its review was still running.

## What

- **`selectPreset` passes the in-memory pair** (`Settings.tsx`). `selectHalf` already hands the new pair to `applyTheme` so a denied `localStorage` write still repaints — but `selectPreset('system')` called `applyTheme` without it. Choose an explicit preset and then System again under a refusing store and the chrome re-read the *stale stored* pair while the pickers still showed the new one. CodeRabbit's finding on #410; regression test `test/dom/settings.test.tsx` "returning to System repaints from the pair the pickers show" (failed with `hive-dark`, expected `nord`, before the fix).
- **The e2e first-stamp helper is strict** (`test/e2e/theme.spec.ts`). It fell back to the live attribute when only one mutation record existed, so a boot script that never stamped at all would have passed on `theme.ts`'s import-time stamp — exactly what the helper exists to catch. It now requires two records with a `null` first `oldValue` and throws otherwise; Chromium does queue a record for the same-value re-stamp, confirmed by the suite.
- **Comments name what is actually duplicated.** "Keep the three in sync" in `theme.ts` / `index.html` named no three things; both now list `THEME_DARK_KEY` / `THEME_LIGHT_KEY` / `DEFAULT_PAIR` / `readHalf` and note the boot script reads only the current scheme's half. `applyTheme`'s comment described one caller; it now describes both.

No behaviour change for a store that accepts writes. `.changesets/system-theme-pair.md` is still unreleased, so this folds into that entry.

## Gates (Windows, Chromium)

`vitest run` 1299/1299 · `biome check` clean on touched files (CRLF-only diffs aside) · `theme.spec.ts` + `settings.spec.ts` 58 passed, 26 snapshot-gated skips.

https://claude.ai/code/session_01T9XS46RXa4dadGnoiNU9Dc


lucascaro (changes requested):
The actual change looks right — `selectPreset` passing the in-memory pair, the strict `firstStamp` helper, and the comment fixes are all good (real diff is ~50 lines across the 5 theme files).

Blocking: the commit converted every touched file to CRLF line endings, so GitHub shows +7631/-7590 across 13 files, including files this PR doesn't mean to change:

- `CHANGELOG.md` (generated) → `block-generated-edits` fails
- `update_apply_{darwin,installable,windows}.go`, `update_preflight_test.go`, `update_shellout_darwin_test.go`, `.changesets/update-refuses-diverged-branch.md`, `docs/design-docs/ui/themes.md` — whole-file rewrites that will conflict with #411
- Biome CI fails on Linux/macOS: "File content differs from formatting output"

Could you re-commit with LF endings and only the 5 theme files (`index.html`, `Settings.tsx`, `theme.ts`, `settings.test.tsx`, `theme.spec.ts`), rebased on main after #411 lands? `git config core.autocrlf input` (or `false`) on the Windows checkout should prevent it recurring. `git diff --ignore-cr-at-eol --stat origin/main` is a quick check that only the intended files remain.

lucascaro:
@qmertesdorf heads-up on line endings: this branch was committed with CRLF. The real change is about 50 lines, but the diff shows +7631/-7590 across 13 files. That includes updater files and `CHANGELOG.md` you never touched, which is why Biome's format check and `block-generated-edits` are red.

#414 adds a repo-level `.gitattributes` (`* text=auto eol=lf`) so git normalizes this on commit from now on. For your local setup on Windows:

```sh
git config --global core.autocrlf input   # commit LF, never convert to CRLF
git config --global core.eol lf
```

Also set your editor to LF for this repo (VS Code: `"files.eol": "\n"`). Once #414 is merged, `git add --renormalize . && git commit` on this branch will strip the CRLF. Please also rebase on `main` afterwards so the stray updater and CHANGELOG diffs drop out.

(Separately, see my question on #413 about whether the refused-write case is real. It affects whether this PR should land at all.)

PR #413, opened 2026-09-14
fix(theme): remember the applied System pair across an OS scheme change

qmertesdorf:
Stacked on #412 (base is its branch; retargets to `main` when it merges). Closes the one limitation #412 left open.

## What

A half picked under a store that refuses writes (a locked-down webview) is applied for the session but never lands in `localStorage`. `initThemeWatch` re-applies `'system'` on an OS scheme change with no pair of its own, so it read the store the write never reached and snapped back to Hive Dark / Hive Light on the next flip — while the Settings pickers, opened fresh, showed the defaults too.

## How

- `theme.ts`: a module-level `livePair` — the pair `applyTheme` was last handed — and `currentPair()`, which returns it when set and the stored pair otherwise. `applyTheme`'s `pair` parameter is now optional; a caller that omits it (boot, the OS-change listener) resolves through `currentPair()`. Null until Settings passes a pair, so boot still reads the store.
- `Settings.tsx`: the pickers open on `currentPair()` instead of `readPair()`, so they show what is painted rather than what the store kept.
- Under a store that accepts writes nothing changes: the stored pair and the live one are the same value.

## Tests

- `theme-apply.test.ts` (fresh module per test via `vi.resetModules`, since this is module state): a later apply without a pair reuses the last one handed; `currentPair()` falls back to storage before any apply and prefers the live pair after.
- `settings.test.tsx`: the pickers open on the applied pair, not the stored one.
- `theme.spec.ts` (e2e): `Storage.prototype.setItem` made to throw, pick Nord / GitHub Light, flip the OS both ways — the chrome follows the session pair and the store still holds nothing.

Each failed first (the e2e snapped back to `hive-light` on the flip).

## Gates (Windows, Chromium)

`vitest run` 1303/1303 · `tsc --noEmit` clean apart from the environmental stale-binding error · `theme.spec.ts` + `settings.spec.ts` 59 passed, 26 snapshot-gated skips.

https://claude.ai/code/session_01T9XS46RXa4dadGnoiNU9Dc


lucascaro:
@qmertesdorf before going further on this (and the refused-write half of #412): have you actually seen `localStorage.setItem` throw in Hive?

I couldn't find a real trigger. Hive runs in WKWebView and WebView2, and both keep localStorage across launches. Nothing we store comes near the quota: the biggest write is the scroll-jump trace snapshot, one key capped at 400 entries. The "store refuses the write" case seems to have started as a defensive test in #410, copying the try/catch "for private mode" pattern that's already in `store.ts`. This PR then adds module-level state (`livePair`) to cover that hypothetical path.

If you did hit it, please share how (OS, build, what failed). That would point to a real bug we should fix at the source rather than work around in the theme code. If it's only defensive, I'd rather close this and keep the theme code simpler. It's also worth noting that the other preferences (density, the update-banner dismissal, and so on) already just don't persist in that case, so theme would be consistent with them.

Holding #412 and #413 until we hear back.

PR #414, opened 2026-09-14
chore: normalize line endings to LF via .gitattributes

lucascaro:
Adds `* text=auto eol=lf` so git normalizes text files to LF on commit, whatever a contributor's `core.autocrlf` is set to.

Why: #412 was committed from a Windows checkout with CRLF endings. That turned a ~50-line change into +7631/-7590 across 13 files, including untouched updater files and `CHANGELOG.md`. It failed Biome's format check and `block-generated-edits`.

- `git add --renormalize .` with this file is a no-op: every tracked text file on `main` is already LF.
- `internal/session/testdata/conformance/.gitattributes` still marks the CRLF terminal fixtures `-text`, and the more specific rule wins, so they're untouched.
- Binary files are auto-detected by `text=auto`.

No user-visible change, so it's labelled `no-changeset`.

## Summary by CodeRabbit

- **Chores**
  - Standardized text file line endings to improve consistency across supported environments.
  - Binary files remain unaffected.


lucascaro:
@qmertesdorf so we don't get all those CRLFs :D

PR #419, opened 2026-09-16
fix: phase 1 follow-ups for spec 416 (rune-safe caps, test shell isolation, changesets README)

lucascaro:
Follow-ups from #417.

- **Rune-safe Text/Target caps**: `applyEventFrame` cut both by byte count. Now it uses `capBytes`, same as Tool/CallID. New test `TestEventModeCapsTargetRuneSafe` fails without the fix. Text was already cleaned downstream by `agentstate.truncate`.
- **`TestTerminalQueriesAreNotWork` failed locally, including on main**: sessions spawn through `$SHELL -l -i`. An interactive rc that prints before the command (seen: a keychain warning plus iTerm shell-integration OSC) made the test take its first sample too early. `freshRegistry` now pins `SHELL=/bin/sh`. CI was unaffected. `scripts/test.sh go` is now green with a zsh login shell.
- **`.changesets/README.md`**: referenced by AGENTS.md, and already excluded by name in release.sh, regen-generated.py and the changesets gate, but it never existed. It now documents the schema that `validate_changeset` enforces.


lucascaro:
Applied `daemon-contract-override`. Claim: no client can tell an old daemon from a new one here. The daemon change only replaces a byte cut with a rune-safe cut on Text/Target, and a new GUI already renders the old daemon's output. The registry change is test-only (`freshRegistry` pins `SHELL`). A bump would restart every user's sessions for no protocol reason.

